PR #182, opened 2014-05-16
Enhancement gif encoding

juliandescottes:


juliandescottes:
@juliandescottesfromfuture 18 months later, this is still not fixed and the comment is still here. I expected more from you!

Issue #249, opened 2015-03-03
Performance issue linked to current colors palette for big sprites


juliandescottes:
For big images (> 200x200), processing frames to extract used colors is time consuming.

Either : 
- deactivate for big sprites
- rely on webworkers for frame processing

juliandescottes:
Fix will be released with v0.5.0

Issue #253, opened 2015-03-07
Issue : Uncaught node.js Error (cannot check current palette in CurrentColorsService)


Dullgarian:
I have this error when I draw and release the mouse button. On Windows 7 64bits.

Uncaught node.js Error 

TypeError: Cannot read property 'id' of null
    at a.CurrentColorsService.isCurrentColorsPaletteSelected_ (file:///C:/Users/MYNAME/AppData/Local/Temp/nw4140_8958/dest/js/piskel-packaged-min-2015-01-23-11-12.js:13:1778)
    at a.CurrentColorsService.onPiskelUpdated_ (file:///C:/Users/MYNAME/AppData/Local/Temp/nw4140_8958/dest/js/piskel-packaged-min-2015-01-23-11-12.js:13:1526)
    at Object.$.event.dispatch (file:///C:/Users/MYNAME/AppData/Local/Temp/nw4140_8958/dest/js/piskel-packaged-min-2015-01-23-11-12.js:2:5581)
    at Object.g.handle.h (file:///C:/Users/MYNAME/AppData/Local/Temp/nw4140_8958/dest/js/piskel-packaged-min-2015-01-23-11-12.js:2:1588)
    at Object.$.event.trigger (file:///C:/Users/MYNAME/AppData/Local/Temp/nw4140_8958/dest/js/piskel-packaged-min-2015-01-23-11-12.js:2:4126)
    at Object.eval (file:///C:/Users/MYNAME/AppData/Local/Temp/nw4140_8958/dest/js/piskel-packaged-min-2015-01-23-11-12.js:2:11963)
    at Function.$.extend.each (file:///C:/Users/MYNAME/AppData/Local/Temp/nw4140_8958/dest/js/piskel-packaged-min-2015-01-23-11-12.js:1:14315)
    at $.fn.$.each (file:///C:/Users/MYNAME/AppData/Local/Temp/nw4140_8958/dest/js/piskel-packaged-min-2015-01-23-11-12.js:1:11034)
    at $.fn.extend.trigger (file:///C:/Users/MYNAME/AppData/Local/Temp/nw4140_8958/dest/js/piskel-packaged-min-2015-01-23-11-12.js:2:11939)
    at Function.a.publish (file:///C:/Users/MYNAME/AppData/Local/Temp/nw4140_8958/dest/js/piskel-packaged-min-2015-01-23-11-12.js:5:844)

juliandescottes:
[...]

PR #259, opened 2015-03-24
Desktop IO Features

leegrey:


leegrey:
This branch adds enhancements to the Desktop build of Piskel, providing new save and load functions using the Node.js FileSystem api.

The new features are currently only active when running in Node-Webkit, although it might be nice to bind the same keyboard shortcuts to the web version.

Keyboard Shortcuts:

ctrl-o to open file
ctrl-shift-s to "save as"
ctrl-s to save

The normal save-to-file ui also uses the new api when running in NodeWebkit, bound to the same behaviour as "ctrl-s". The desktop version could perhaps do with a modified ui, with a button for save-as. I also thought that the desktop version might always save into the local storage also, using the save path as an id. The file-recovery feature is nice if the app crashes for some reason.

The savePath is stored in the Piskel instance, and propagated when a new instance is created. (Except when a new document is opened.) These are the events I've covered:

savePath stored on load
savePath stored on drag-to-open
savePath propagated on redo / undo
savePath propagated on resize

I've used the app for a few nights and have not noticed any other situations where it loses the savePath. Can you think of any I've missed? I notice that there doesn't seem to be a "new Piskel" button anywhere, but that would be another situation to deal with.

The '.piskel' extension is automatically added to filename if the user does not type it. I've realised that this is potentially dangerous. For example, say you type "foo" but there is already a file called "foo.piskel", you will not be warned that it already exists, and it will just be overwritten. Since this kind of checking behaviour is not supported through the basic dialog (afaik), some custom dialog would have to be used. 

Also, I get tons of warnings during compile (mostly from files I have not modified) and must use grunt --force to build.

leegrey:
Ah, I see it has failed the Travis Cl build. The issue here is that I'm using checking for the existence of 'process' and 'require' to detect NodeWebkit, but they are undefined as far as the compiler is concerned. (I'm checking to see *if* they are undefined.) I can probably suppress the warning by doing something like window["process"] and window["require"].

leegrey:
Ok It complained about that as well. Any ideas?

juliandescottes:
I usually use window.require and window.process. 
I think it first complained about the unnown variables and then about the bracket notation.

Then we could add them to the jshint configuration, in Gruntfile.js ?

juliandescottes:
Just checked the Travis CI build, it failed because of the Closure Compiler step. 
I should get rid of this, initially we wanted to rely on closure annotations but never really used it.

If you want to make Closure Compiler recognize the new globals, you can edit build\closure\piskel-closure-externs.js

leegrey:
Oh, right. I'll try that tonight. Thanks. BTW I'm in New Zealand, so I think we're at exactly opposite time zones. (You are in France, right?)

juliandescottes:
Yes I'm in France :) Good morning then !

Regarding the compilation warnings you see, they are linked to the Closure Compiler, they don't block the build.

However, the steps that can fail the build are : 
- jshint
- leading indent (checking indentation is 2-space in all JS files)
- karma runner running Jasmine unit tests
- casper js integration tests with a headless browser

Let me know if one of those is giving you a hard time.
The build should pass without using --force :)

juliandescottes:
After renaming require & process to window.require and process, the build is OK for your branch on my machine.

I tried your branch and the new feature works really well !

I think we should update at least two things : 
- add the filename to the "Successfully saved" notification
- change the "Save Offline As File" section of the save panel to also display the filename that is used when saving (maybe display the whole savepath). Instead of 'Your piskel will be downloaded as New Piskel-20140727-104433.piskel' display either nothing if the savepath is empty or 'Saving as [savepath]'

Is that ok with you ?

As you suggested, it would be nice to have save & save as button, and maybe tune the Save UI a bit differently when used in desktop mode. But I think this can be done later.

leegrey:
Those changes sound good to me. Are you working on those or did you want me to do them? I have some other work I need to focus on soon, so I won't be able to do too much for a while, but I will do what I can to put the final touches on this feature. I'm excited to see the update in the public release. :)

juliandescottes:
Congratulations on getting Travis back to life \o/

> Are you working on those or did you want me to do them?

I'm fine with both, we can merge your PR as it is and open issues to track the cosmetic changes.

I need to find some time to fix #249 and #253 before doing the new version + testing.
Given my current schedule, a new version could be ready for beginning of April. That leaves some time for you or me to improve the save panel.

Let me know if I should go ahead with the merge or if you want to spend some more time on it before.

juliandescottes:
Another question :) 

I tested your build on Mac OS, I don't have access to a Windows computer for a few days. 
Did you test on Windows ?

leegrey:
I ran 'grunt nodewebkit' and tested the release build on windows. Everything seems to be working fine.

I should have a chance to look at those ui changes within the next week. Let me know if you get to them first.

juliandescottes:
Great ! I'll leave it to you, next week looks good.

Do you want to keep this PR for the UI changes or should we merge before ?
(I'll do a last review review before merging)

leegrey:
May as well go ahead and merge now.

I will try to get to those UI changes asap, but probably not until tomorrow night at the earliest.

leegrey:
I made those UI changes, and also hid the "Save to browser" part of the ui when running on desktop.

src/js/service/DesktopStorageService.js
```diff
@@ -0,0 +1,71 @@
...
+    });
+  };
+
+  ns.DesktopStorageService.prototype.savePiskelAs = function (savePath) {
+    var serialized = this.piskelController.serialize();
+    // TODO: if there is already a file path, use it for the dialog's
+    // working directory and filename
+    pskl.utils.FileUtilsDesktop.saveAs(serialized, null, 'piskel', function (selectedSavePath) {
+      this.onSaveSuccess_();
+      this.piskelController.setSavePath(selectedSavePath);
```
juliandescottes: L55 and L56 should be swapped, since we need the savePath inside the onSaveSuccess_

Will change it after the merge.

juliandescottes:
OK looks good to me ! Merging. :shipit: 

(now I will merge your changes with my current branch, where I refactored quite a bit the settings controllers, that will be fun :P )

Thanks a lot for your contribution ! As I said, release will probably be ready beginning of April. I'll make sure to highlight your involvement.

And if you want to add/fix anything else, you're more than welcome.

leegrey:
Thanks! It was fun to contribute. Piskel is really nice software, I'm glad I could help out :)

PR #275, opened 2015-04-29
Zip-Export: Split by layers option

MaxEden:
Added option to export non-merged frames

### DIFF
--- a/src/js/controller/settings/exportimage/PngExportController.js
+++ b/src/js/controller/settings/exportimage/PngExportController.js
@@ -13,6 +13,8 @@
     this.pngFilePrefixInput = document.getElementById('zip-prefix-name');
     this.pngFilePrefixInput.value = 'sprite_';
 
+    this.splitByLayersCheckbox = document.getElementById('zip-split-by-layers');
+
     var downloadButton = document.querySelector('.png-download-button');
     this.addEventListener(downloadButton, 'click', this.onPngDownloadButtonClick_);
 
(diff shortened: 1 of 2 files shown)


juliandescottes:
Hi, 

Thanks for the PR, I'll have a look later today !

juliandescottes:
@MaxEden Can you give me a bit of context here ? How do you take advantage of this splitted export in your workflow ?

Thanks !

MaxEden:
I draw sprites for games, and the result sprites often need to be made of several separate parts, for example: legs/torso/arms/weapon. It's easy to animate the pieces together, but I still need them separated so I can combine them dynamically or apply sfx only on some parts.
https://www.dropbox.com/s/mgbce3ztho2nsug/Squirrel.piskel

juliandescottes:
Ok ! Thanks for info, seems reasonable :)

src/js/controller/settings/exportimage/PngExportController.js
```diff
@@ -30,12 +32,34 @@
   ns.PngExportController.prototype.onZipButtonClick_ = function () {
     var zip = new window.JSZip();
 
-    for (var i = 0 ; i < this.piskelController.getFrameCount() ; i++) {
-      var frame = this.piskelController.getFrameAt(i);
-      var canvas = this.getFrameAsCanvas_(frame);
-      var basename = this.pngFilePrefixInput.value;
-      var filename =  basename + (i + 1) + '.png';
-      zip.file(filename, pskl.utils.CanvasUtils.getBase64FromCanvas(canvas) + '\n', {base64: true});
+    function splittedExport() {
```
juliandescottes: We don't normally use named functions in Piskel, but try to attach functions to prototypes (except for some anonymous functions in forEach constructs, callbacks and so on ...).

To be consistent, we should do the same here.

src/js/controller/settings/exportimage/PngExportController.js
```diff
@@ -30,12 +32,34 @@
...
+          var basename = this.pngFilePrefixInput.value;
+          var filename = 'l' + j + '_' + basename + (i + 1) + '.png';
+          zip.file(filename, pskl.utils.CanvasUtils.getBase64FromCanvas(canvas) + '\n', {base64: true});
+        }
+      }
+    }
+
+    function mergedExport() {
+      for (var i = 0; i < this.piskelController.getFrameCount(); i++) {
+        var frame = this.piskelController.getFrameAt(i);
```
juliandescottes: We could mutualize some code between mergedExport and splittedExport ?

MaxEden: Maybe, but not much without drenching it with ifs. Perhaps I don't see how.

src/js/controller/settings/exportimage/PngExportController.js
MaxEden: Something like this?
http://pastebin.com/gGZCmAyc

juliandescottes: Yes , that'd be great

src/js/controller/settings/exportimage/PngExportController.js
juliandescottes: Fine with keeping this as it is.

juliandescottes:
@MaxEden : if you can update your PR with the change you linked on pastebin, I'll merge it right after.

Thanks !

MaxEden:
Updated btw

juliandescottes:
Thanks for the ping, didn't see that.

PR #284, opened 2015-06-01
Adding spritesheet import - fixes #188

### LINKED ISSUE #188: Enhancement: Import existing spritesheet
Would love to be able to import my existing sprite sheets into Piskel by defining not only the frame size, but the number or rows/cols.

JALissiak:
- Updated the import dialog to allow users to specify the number of frames in the source image (which defaults to 1 x and 1 y)
- Setting the frame count for x and y will draw a dotted line in the preview that shows where the image will be split into individual frames
- Changing the frame count will also auto adjust the image width/height so that it displays the size of an individual frame, which can then be manually changed to resize the resultant frames into the desired final size
- When imported with a frame count above 1, the source image will be split into the different frames and loaded just as if it were an animated gif
- This allows users to import existing spritesheet pngs, including those produced by the piskel export function (additional work will be needed to specify the frame layout on export - right now a 'square' layout as shown in the image below will still be exported as a single horizontal strip)

image

### DIFF
--- a/src/css/dialogs-import-image.css
+++ b/src/css/dialogs-import-image.css
@@ -2,28 +2,51 @@
 /* Import dialog                                                                                 */
 /************************************************************************************************/
 
+.import-subsection {
+  margin-left: 25px;
+}
+
+.import-section:not(.import-subsection) > .dialog-section-title {
+  width: 50px;
+}
+
+.import-section-preview-title {
+  position: absolute;
+  margin-left: 50%;
+  margin-top: -28px;
+}
+
 .import-section-preview {
-  position: relative;
+  position: absolute;
   display: inline-block;
-  height: 60px;
-  width: 60px;
   border: 1px dashed #999;
   border-radius: 3px;
+  margin-left: 50%;
+}
+
+.import-section-preview img {
+  max-width: 220px;
+  max-height: 220px;
 }
 
 .import-section-preview.no-border {
-    border-color: transparent;
+  border-color: transparent;
 }
 
 .import-section-preview canvas {
-    position: absolute;
-    left: 0;
-    top: 0;
+  position: absolute;
+  left: 0;
+  top: 0;
 }
 
 .dialog-section-title {
   display : inline-block;
-  width: 70px;
+  width: 80px;
+}
+
+.dialog-section-radio {
+  margin-top: 15px;
+  vertical-align: sub;
 }
 
 .import-size-field:nth-of-type(2) {
@@ -34,12 +57,14 @@
   display: inline-block;
   overflow: hidden;
 
-  height: 2rem;
   word-break : break-all;
   vertical-align: middle;
   font-style: italic;
   font-weight: normal;
   text-shadow: none;
+  width: 200px;
+  white-space: nowrap;
+  text-overflow: ellipsis;
 }
 
 [name=smooth-resize-checkbox] {
@@ -48,11 +73,12 @@
 
 .dialog-import-body {
   padding:10px 20px;
-  font-size:1.5em
+  font-size:1.3em
 }
 
 .import-button {
   font-size: 1em;
   height: auto;
   padding: 5px 10px;
+  margin-top: 15px;
 }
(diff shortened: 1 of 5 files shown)


juliandescottes:
Thanks a lot for your contribution. The code change looks good to me, that's great !

I tried out the feature and I just have one comment : many spritesheets I find have some padding on the right/bottom of the image.
Therefore when you split by a X rows, Y columns, it's sometimes impossible to properly isolate the frames. (see http://www.swingswingsubmarine.com/wp-content/uploads/2010/11/dude_animation_sheet.png for instance)

I was wondering if we shouldn't go the other way around, and ask for a frame size. User could specify the expected frame width/height, and adjust it using the grid preview you developed. We would then ignore empty/incomplete frames during the import.

But maybe I was just unlucky with the spritesheets I found ?
What do you think ?

JALissiak:
You are right, I guess a lot of spritesheets are packed into images that are powers of 2 in size, which is fine until the actual frame size is not fully divisible by that, as with your example. Mostly I tend to work with frame sizes that are powers of 2 themselves (16, 32, etc) as I know I'm going to be packing them together, so I didn't think about it.

I guess the better approach is to (as you say) allow to user to specify the frame size, and a possible offset value, and only split into frames where it can fit a full frame of that size (leaving out the gap at the right/bottom).

What are your thoughts on the UI for that? Add a new checkbox for 'split into frames' and if checked, reuse the 'Size' as frame size, and have what I've currently got as 'Frames' to be 'Offset' instead?
Or maybe have a completely new set of 'frame size' and 'frame offset' inputs, so that the user can use the existing 'Size' input to specify the output frame size in case they want to resize the resulting split frames?
It might also be useful to make the preview image bigger so its easier to see where the frames are split, but I couldn't figure out a good way to do that and still leave space for all the options we need.

I could potentially add a checkbox for 'ignore empty frames' in case people want to keep the empty ones for some reason, but maybe that is overkill.

I'll take a look later and see what I can come up with. But any ideas you have would be great.

juliandescottes:
Ultimately I think the image import and spritesheet import could be really separated.
But in the meantime, a suggestion could be :



- move the preview to the right, give it more space (screenshot : at 220px max-width & max height)
- reduce the popup font size (screenshot : down from 1.5em to 1.3em)
- add a radio to choose between image import and spritesheet import
- increase a bit the popup size (screenshot : 550x350)

What do you have in mind for the offset ? If there is padding on the top/left of the image, and not only on the bottom right ?

> I could potentially add a checkbox for 'ignore empty frames' in case people want to keep the empty ones for some reason, but maybe that is overkill.

I tend to agree, it's an overkill, let's just skip the empty or incomplete frames.

JALissiak:
I've updated the code so that you can now specify the size and offset as you suggested. The new UI looks like this:
image

The offset will be useful if a spritesheet is packed into a larger image but does not start at the top left.
 
Also the 'blank' frames are ignored by comparing the canvas's dataUrl to a empty canvas, which will only work if the background color of the spritesheet is transparent. I did this as its faster than comparing all the pixels in each frame to see if it contains anything. This seems fine to me as the workaround is for a user to just delete empty frames themselves after they have imported it if they use a non-transparent background color. But let me know if you have a better idea.

JALissiak:
Also - Thanks for the great tool, I love this thing :)

juliandescottes:
That looks really great ! :+1: 

Made a few comments in the review, if you still have energy to spare :) It should be ready to merge after that.

:bulb:  Food for thoughts : even with the grid, it can be difficult to spot the "good" sprite size. Any ideas on how we could make this easier ? Maybe a live preview of the import ?

> Also - Thanks for the great tool, I love this thing :)

That's really nice to hear ! Thanks a lot for your contribution :)

JALissiak:
Thanks. I've updated the code based on your feedback.

A live preview would be a nice addition. I guess the workaround right now is to just import again until you get it right? So not too bad I hope. Shouldn't be an issue for sheets you've made yourself though, as you already know the size.

juliandescottes:
Looks good to me, merging !
I will try to release a new version and upload it to piskelapp.com this weekend.

(Do you have a twitter account ? I'd like to mention you when announcing the release on twitter, if you're ok?)

JALissiak:
I don't really use Twitter. Feel free to say "github user JALissiak" or something. But it's not necessary. Just glad I could be useful :)

Issue #290, opened 2015-07-21
Move & Selection tool


AlexHolly:
Is there a way to move a Selection without cut and paste it?
After a selection i can onlye move the selection box without the pixels inside.

When i switch to the selection tool i have sometimes to click once in the draw area to get the selection tool to work.Is it a bug?

juliandescottes:
> Is there a way to move a Selection without cut and paste it?

That's the only way currently. There are other issues requesting move/selection improvements.

> When i switch to the selection tool i have sometimes to click once in the draw area to get the selection tool to work.Is it a bug?

That's a bug !

TheThirdOne:
@juliandescottes, would you be open to a PR for changing the move tool so that when it is clicked on top of a selection it will only move the selection?

juliandescottes:
@TheThirdOne sure ! Having the selection persistent between tools will be really cool. Could then use it for the paint bucket.

Right now we have one 'overlay' frame which we use to draw all the temporary pixels, before committing them to the real frame. Each tool is responsible for drawing the overlay frame. 

I think the easiest will be to stop using this overlay frame for the selection. We can either use a similar overlay frame, or even switch to a completely different display, using a DOM element here. Using another frame will probably be faster for a first version.

TheThirdOne:
@juliandescottes, I got it working, but I have a few questions before I submit a PR.

1. Why does pskl.selection.BaseSelection exist when you could use pskl.tools.drawing.selection.BaseSelection? I had to move a function over from pskl.tools.drawing.selection.BaseSelection into pskl.selection.BaseSelection to make   it work.
2. What do you think about leaving the selection around the entire time? With your current code the selection is deleted immediately; I had to modify it to leave the selection in place. Should the selection stay visible? Should there be a way to delete it?
3. I'm not sure how ctrl-z works, but for some reason it works correctly which surprises me.

juliandescottes:
> 1. Why does pskl.selection.BaseSelection exist

pskl.tools.drawing.selection.BaseSelect is the tool : what should be done when user moves the tool, clicks, starts dragging etc ...
pskl.selection.BaseSelection is the object that represents selected pixels and has some utility functions associated to it. The bare minimum is that the selection should hold the selected pixels. Utility functions can be reviewed one by one, no worries.

Before moving a method out of the Selection class, the only question to ask is probably whether this could be called from the selectionManager.

(That being said you should also be able to call selectionTool.selection.yourMethod() ?)

[...]

PR #299, opened 2015-08-29
Added moving selection [WIP]

TheThirdOne:
PR for #290

There are two main problems before this can be merged in:
1. The selection currently does not stay visible after changing to another tool.
2. Pressing escape doesn't remove the selection.

### DIFF
--- a/src/js/selection/BaseSelection.js
+++ b/src/js/selection/BaseSelection.js
@@ -31,4 +31,10 @@
 
     this.hasPastedContent = true;
   };
+  
+   ns.BaseSelection.prototype.isInSelection = function (col, row) {
+    return this.pixels.some(function (pixel) {
+      return pixel.col === col && pixel.row === row;
+    });
+  };
 })();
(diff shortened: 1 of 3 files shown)


juliandescottes:
Thanks ! Looks like a good start.

Some comments on top of what you mentioned : 

1. When a selection is active, the move tool should always move the selection, regardless of where you clicked for the move. If there is no selection, the move tool should move the whole frame. (simply )

2. When using the moveTool with a selection, the selection should also be moved.

3. The selection is not persisted during when saving a move state. I think the easiest is to backup the pixels of the selection when saving the move state and use this information when replaying the state. Since the move tool is the only tool impacted by the selection for now I think this should be fine.

What do you think ?

TheThirdOne:
1. I can make that happen as soon as dismissing selection works.
2. Yup
3. That will probably be the last change I make.

juliandescottes:
Sounds good, let me know if you need info/help on this !

juliandescottes:
Linked issue closed, closing the PR

Issue #247, opened 2015-02-07
Export spritesheets with sizes that constrain to powers of 2


LuckySvenn:
Hey, I really like your application!

But it would be really nice to maybe have a checkmark box that could change the width and height sizes of the spritesheet to only powers of 2.

So, let's say I've got a spritesheet that's about to be exported that is currently 160x32. It'd be nice to check a checkmark box that does something like changes 160 to 256.

I work with engines that require textures that constrain to the power of 2 for rendering quality, and this would be really useful for me!

Thanks!

juliandescottes:
Let's say we have 4 frames @ 40x32 => spritesheet of 160x32.
  - should it be resized (scaled up) to 256*51 ? 
  - should it be 256*32, with something to fill the empty space ?

(What is the name of the engine you are using ?)

LuckySvenn:
The latter implementation.

I'm using libGDX, and in their [wiki](https://github.com/libgdx/libgdx/wiki/Spritebatch%2C-Textureregions%2C-and-Sprites) under the 'Texture' header, it suggest using textures with dimensions that are powers of 2.

juliandescottes:
Will be fixed with #473, closing as duplicate.

PR #300, opened 2015-08-31
Adds the ability to export a scaled spritesheet.

jnlopar:
This is a patch for the second feature I requested in this tweet: https://twitter.com/lopar/status/637802318008717312

Please let me know if there are any tweaks to the UX or whatnot. For one, I was debating whether the scale should appear as a general export setting or as part of the spritesheet setting section, since it currently only works for spritesheets. I went with general export since it'd be nice to extend it to others types such as ZIP in the future, but I'll defer to whatever you think is best there.

The scaling method with canvasses should work on all modern browsers (IE 11+ and Edge, Chrome, Firefox, Safari, Opera probably).

### DIFF
--- a/src/js/controller/settings/exportimage/ImageExportController.js
+++ b/src/js/controller/settings/exportimage/ImageExportController.js
@@ -14,6 +14,7 @@
     var scalingFactorInput = document.querySelector('.scaling-factor-input');
     scalingFactorInput.value = pskl.UserSettings.get(pskl.UserSettings.EXPORT_SCALING);
     this.addEventListener(scalingFactorInput, 'change', this.onScalingFactorChange_);
+    this.addEventListener(scalingFactorInput, 'input', this.onScalingFactorChange_);
     this.updateScalingFactorText_(scalingFactorInput.value);
 
     this.pngExportController.init();
--- a/src/js/controller/settings/exportimage/ImageExportController.js
+++ b/src/js/controller/settings/exportimage/ImageExportController.js
@@ -24,6 +24,7 @@
   ns.ImageExportController.prototype.destroy = function () {
     this.pngExportController.destroy();
     this.gifExportController.destroy();
+    this.superclass.destroy.call(this);
   };
 
   ns.ImageExportController.prototype.onScalingFactorChange_ = function (evt) {
--- a/src/js/controller/settings/exportimage/ImageExportController.js
+++ b/src/js/controller/settings/exportimage/ImageExportController.js
@@ -7,7 +7,15 @@
     this.gifExportController = new ns.GifExportController(piskelController);
   };
 
+  pskl.utils.inherit(ns.ImageExportController, pskl.controller.settings.AbstractSettingController);
+
   ns.ImageExportController.prototype.init = function () {
+    // Output Scaling Factor
+    var scalingFactorInput = document.querySelector('.scaling-factor-input');
+    scalingFactorInput.value = pskl.UserSettings.get(pskl.UserSettings.EXPORT_SCALING);
+    this.addEventListener(scalingFactorInput, 'change', this.onScalingFactorChange_);
+    this.updateScalingFactorText_(scalingFactorInput.value);
+
     this.pngExportController.init();
     this.gifExportController.init();
   };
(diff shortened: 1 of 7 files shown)


src/js/utils/BlobUtils.js
```diff
@@ -23,9 +23,28 @@
...
+        ctx.mozImageSmoothingEnabled = false;
+        ctx.webkitImageSmoothingEnabled = false;
+        ctx.msImageSmoothingEnabled = false;
+        ctx.imageSmoothingEnabled = false;
+
+        ctx.drawImage(sourceCanvas, 0, 0, sourceCanvas.width, sourceCanvas.height, 0, 0, canvas.width, canvas.height);
+      } else {
+        canvas = sourceCanvas;
+      }
+
```
juliandescottes: You can reuse ImageResizer.resize to resize the canvas. The resizing should also be moved to the caller (ie PngExportController). CanvasToBlob is a helper and can not assume it is used for exporting.

src/templates/settings/export.html
```diff
@@ -1,4 +1,14 @@
 <div class="settings-section">
+
+  <div class="settings-title settings-section--export">
+    General Export Settings
+  </div>
+  <div class="settings-item">
+    <span class="settings-description" style="display:block"><label for="scaling-factor">Output Scaling Factor</label></span>
+    <span class="settings-description" style="display:block">Currently affects only spritesheet export.</span>
```
juliandescottes: I know you just applied what you saw a few lines above, but can I ask you to do a bit of extra cleanup ? :) 

Change the spans style=display:block to divs (including the first one in the file you didn't change) and remove the display inline-block from .settings-description rules in settings.css ?

If you move this to the PNG section you can also get rid of the second part of the description.

juliandescottes:
Hi ! (and thanks a lot for contributing !)

Let's start with a user setting for now, but the spritesheet export needs to go have a dedicated dialog to take care of scaling and other features such as issues #213 and #247. This dialog would support ZIP, PNG with all the options needed. As you've probably seen, the drawer is nice and all, but you run out of space pretty quickly.

Now I'm fine fine with having a temporary solution without the dialog :). I would move this setting to the PNG export section though. I think the feature will be easier to understand this way.

I added some comments in the files review, if you want to have a look.

Nice work, thanks again :)

jnlopar:
Thanks for the feedback! Sorry it took me a couple days to have time to sit down with this and update. Always happy to do cleanup on a codebase where needed. :) ImageResizer is definitely a cleaner approach as well.

juliandescottes:
Sorry for the delay ! The change looks good, ready to merge. And thanks for the cleanup :) 

I'll just test it one last time this weekend.

Thanks !

juliandescottes:
Looks really good ! 

Maybe one last thing before merging : I would like the UI to be a bit more compact. Move the label inside the .scaling-factor div, right before the input range. Label can be changed to "Scale" to be shorter. 

If you want you can add tooltips to any element following the exanple below : 
    
    <div class="scaling-factor" 
      title="Scale the exported PNG spritesheet"
      rel="tooltip" 
      data-placement="top"
    >

Other than this it's ready to merge.

src/js/controller/settings/exportimage/ImageExportController.js
```diff
@@ -7,7 +7,15 @@
     this.gifExportController = new ns.GifExportController(piskelController);
   };
 
+  pskl.utils.inherit(ns.ImageExportController, pskl.controller.settings.AbstractSettingController);
+
   ns.ImageExportController.prototype.init = function () {
+    // Output Scaling Factor
+    var scalingFactorInput = document.querySelector('.scaling-factor-input');
+    scalingFactorInput.value = pskl.UserSettings.get(pskl.UserSettings.EXPORT_SCALING);
+    this.addEventListener(scalingFactorInput, 'change', this.onScalingFactorChange_);
```
juliandescottes: It would be nice to have a listener on the 'input' event as well : 
```javascript
    this.addEventListener(scalingFactorInput, 'change', this.onScalingFactorChange_);
    this.addEventListener(scalingFactorInput, 'input', this.onScalingFactorChange_);
```
On browsers that support the 'input' event properly on range inputs, this will fire everytime the value is updated, so you will see the number change in real time.

juliandescottes: We need to make sure the event listeners added are removed when the panel is closed. This is handled by  pskl.controller.settings.AbstractSettingController#destroy . The PngExportController and GifExportController both extend AbstractSettingController without overriding the destroy method. All their events are automatically cleaned up.

Here you can either add a call to ```this.superclass.destroy.call(this);``` at the end of the destroy method. Or you can move the event's logic to the PngExportController.

juliandescottes:
Added 2 small comments on ImageExportController, would be great if we could also take care of this before merging :)

Thanks !

src/js/controller/settings/exportimage/ImageExportController.js
jnlopar: Done, went with the superclass call for now.

jnlopar:
Sorry for the long delay again, I think I've addressed the last few comments. :)

juliandescottes:
Looks good to merge !

Thanks a lot for your contribution. For information, I will try to release a new version of Piskel either this week or next week. Your feature will then be available on piskelapp.com and in the new desktop applications. 

I'll make sure to mention you when I publish the release :) (and I should have a contributors' section somewhere, but not sure I'll get around to do that this week)

Thanks again !

PR #308, opened 2015-09-14
Introduce basic dithering tool

grosbouddha:
image


captainbrosset:
:+1:

juliandescottes:
Great time saver, thanks a lot for this. Added a few suggestions in the comments.

The last one regarding the PaletteController cleanup can be done in a separate ticket, let me know how you feel about it.

src/js/tools/drawing/DitheringTool.js
```diff
@@ -0,0 +1,31 @@
...
+
+  /**
+   * @override
+   */
+  ns.DitheringTool.prototype.applyToolAt = function(col, row, color, frame, overlay, event) {
+    // Use primary selected color on cell with either an odd col or row.
+    // Use secondary color otherwise.
+    // When using the right mouse button, invert the above behavior to allow quick corrections.
+    var usePrimaryColor = (col + row) % 2;
+    var invertColors = event.button === Constants.RIGHT_BUTTON;
```
juliandescottes: We cannot rely on event.button here, because applyToolAt can be triggered by a mousemove event, which will not have the expected 'button' value. (see comment in DrawingController getCurrentColor_).

The only way to know if we are on the right button is to check the color here. Otherwise we could modify the event object passed by the DrawingController to the tools

grosbouddha: I added a mousedown check in the dithering#applyToolAt method.

juliandescottes:
Thanks for addressing the comments. Merging.

PR #309, opened 2015-09-15
Dithering tests

grosbouddha:
Add unit tests for SelectedColorsService and drawing tests for the dithering tool.


juliandescottes:
Looks good to me !

PR #310, opened 2015-09-16
Integrate selected colors service

grosbouddha:
The goal of this list of changes is to remove the color arguments from all BaseTool/BaseSelect interfaces and rely on centralized/shared services to provide colors that tools should use.

The color that a tool should use is based on the current primary/secondary colors and if the right/left button of the mouse is used.
SelectedColorsService provides the primary/secondary colors.
The new service MouseStateService provides the mouse button state (and encapsulate properly the mousemove mouse button type bug in FF/IE).

Each tool can use these two services in the getToolColor base method to know which color to use, and the color argument can be removed from the applyToolAt/releaseToolAt/... methods.

### DIFF
--- a/src/js/tools/drawing/ColorPicker.js
+++ b/src/js/tools/drawing/ColorPicker.js
@@ -16,7 +16,7 @@
   /**
    * @override
    */
-  ns.ColorPicker.prototype.applyToolAt = function(col, row, color_legacy, frame, overlay, event) {
+  ns.ColorPicker.prototype.applyToolAt = function(col, row, frame, overlay, event) {
     if (frame.containsPixel(col, row)) {
       var sampledColor = frame.getPixel(col, row);
       if (event.button == Constants.LEFT_BUTTON) {
--- a/src/js/tools/drawing/ColorPicker.js
+++ b/src/js/tools/drawing/ColorPicker.js
@@ -16,7 +16,7 @@
   /**
    * @override
    */
-  ns.ColorPicker.prototype.applyToolAt = function(col, row, color, frame, overlay, event) {
+  ns.ColorPicker.prototype.applyToolAt = function(col, row, color_legacy, frame, overlay, event) {
     if (frame.containsPixel(col, row)) {
       var sampledColor = frame.getPixel(col, row);
       if (event.button == Constants.LEFT_BUTTON) {
--- a/src/js/tools/drawing/ColorPicker.js
+++ b/src/js/tools/drawing/ColorPicker.js
@@ -19,9 +19,9 @@
   ns.ColorPicker.prototype.applyToolAt = function(col, row, frame, overlay, event) {
     if (frame.containsPixel(col, row)) {
       var sampledColor = frame.getPixel(col, row);
-      if (event.button == Constants.LEFT_BUTTON) {
+      if (pskl.app.mouseStateService.isLeftButtonPressed()) {
         $.publish(Events.SELECT_PRIMARY_COLOR, [sampledColor]);
-      } else if (event.button == Constants.RIGHT_BUTTON) {
+      } else if (pskl.app.mouseStateService.isRightButtonPressed()) {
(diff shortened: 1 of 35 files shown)


src/js/tools/drawing/Rectangle.js
```diff
@@ -16,11 +16,14 @@
...
-  ns.Rectangle.prototype.draw_ = function (col, row, color, targetFrame) {
+  /**
+   * @override
+   */
+  ns.Rectangle.prototype.draw = function (col, row, color, targetFrame) {
     var strokePoints = pskl.PixelUtils.getBoundRectanglePixels(this.startCol, this.startRow, col, row);
     for (var i = 0 ; i < strokePoints.length ; i++) {
       // Change model:
-      targetFrame.setPixel(strokePoints[i].col, strokePoints[i].row, color);
+      targetFrame.setPixel(strokePoints[i].col, strokePoints[i].row, this.getToolColor());
```
juliandescottes: I think you should use "color" instead of getToolColor().
The logic to swap TRANSPARENT_COLOR for the SELECTION_TRANSPARENT_COLOR is done in the ShapeTool here. It allows to preview your rectangle if you are using the transparent color.

src/js/tools/drawing/Lighten.js
```diff
@@ -41,19 +41,24 @@
   /**
    * @Override
    */
-  ns.Lighten.prototype.applyToolAt = function(col, row, color, frame, overlay, event, mouseButton) {
+  ns.Lighten.prototype.applyToolAt = function(col, row, frame, overlay, event, mouseButton) {
```
juliandescottes: (not related to your change) We should remove this unused mouseButton argument. Up to you if you want to do it with this PR or separately.

src/js/tools/drawing/Lighten.js
```diff
@@ -65,9 +70,15 @@
       }
     }
     if (color) {
```
juliandescottes: After checking the code you can get rid of the check : tinycolor and tinycolor.darken/lighten always return a valid tinycolor instance.

src/js/tools/drawing/Lighten.js
```diff
@@ -65,9 +70,15 @@
...
-      usedPixels[key] = true;
-      this.superclass.applyToolAt.call(this, col, row, color.toRgbString(), frame, overlay, event);
+      // Convert tinycolor color to string format.
+      color = color.toRgbString();
+    } else {
+      // Not sure why this check exists in the first place.
```
juliandescottes: After checking the code you can get rid of the check : tinycolor and tinycolor.darken/lighten always return a valid tinycolor instance.

src/js/tools/drawing/ColorPicker.js
```diff
@@ -16,7 +16,7 @@
   /**
    * @override
    */
-  ns.ColorPicker.prototype.applyToolAt = function(col, row, color, frame, overlay, event) {
+  ns.ColorPicker.prototype.applyToolAt = function(col, row, frame, overlay, event) {
```
juliandescottes: Use the MouseStateService here ?

juliandescottes:
Thanks for the refactor ! 

Good to merge after addressing my comments. 
(no need for re-review, I think you have the permissions to merge to master ?)

src/js/tools/drawing/Lighten.js
grosbouddha: Separately.

src/js/tools/drawing/ColorPicker.js
grosbouddha: Sure, good catch.

src/js/tools/drawing/Lighten.js
grosbouddha: done

src/js/tools/drawing/Rectangle.js
grosbouddha: done

PR #326, opened 2015-11-03
Tablet Touch & s-pen

AlexHolly:
Needs a little bit more testing.
I had problems pushing it but it was looking good.

I was only able to create a pixel with "down & up" now dragging also works.


src/js/controller/DrawingController.js
```diff
@@ -140,6 +142,12 @@
     $.publish(Events.MOUSE_EVENT, [event, this]);
     var frame = this.piskelController.getCurrentFrame();
     var coords = this.getSpriteCoordinates(event.clientX, event.clientY);
+    if (typeof event.targetTouches != 'undefined' &&
+        typeof event.targetTouches[0] != 'undefined' &&
+        typeof event.targetTouches[0].pageX != 'undefined' &&
+        typeof event.targetTouches[0].pageY != 'undefined') {
```
juliandescottes: Might be enough to simply check if the objects are not falsy ````if(event.targetTouches && event.targetTouches[0] ...)```` and skip the checks on pageX and pageY, since targetTouches contains Touch instances.

src/js/controller/DrawingController.js
```diff
@@ -252,6 +267,12 @@
   ns.DrawingController.prototype.onMouseup_ = function (event) {
     var frame = this.piskelController.getCurrentFrame();
     var coords = this.getSpriteCoordinates(event.clientX, event.clientY);
+    if (typeof event.changedTouches != 'undefined' &&
+        typeof event.changedTouches[0] != 'undefined' &&
+        typeof event.changedTouches[0].pageX != 'undefined' &&
+        typeof event.changedTouches[0].pageY != 'undefined') {
```
juliandescottes: Maybe we could use changedTouches as well in the other two event handlers ? 

From reading the spec and quick testing I think it should give us the relevant information for the touchstart and touchmove events as well.Then we could mutualize the check done in those if in a helper function (hasChangedTouchPoint(event) ?)

src/js/controller/DrawingController.js
juliandescottes: (just a question, no need to change anything here) 

Is there a good reason to prefer pageX/pageY to clientX/clientY here ? (In theory for Piskel we clientX/Y and pageX/Y should be equivalent because we have no scrollbar right ?)

juliandescottes:
Thanks a lot for the PR ! I tested your changes (on an emulator since I don't have a tablet :) ) and it seems to work well ! 

I have a few comments on the review if you can have a look.

After that we'll be good to merge and it will be released with 0.6.0 (still don't know when though !)

AlexHolly:
I don't like that many checks either but it was the only way to get it working. I'm really not into js so I will check from top to bottom of you posts.

1. if (typeof event.targetTouches && ...
Moving the Mouse -> Error: TypeError: a.targetTouches is undefined

2. changed pageX/Y to clientX/Y -> works

3. removing the clientX/Y checks -> works

4. changedTouches on all event handlers -> works
  I don't know the difference between changedTouches and targetTouches I had to use changedTouches on one place because targetTouches didn't work.

5. hasChangedTouchPoint - Maybe you add this on your own I don't really get it and you are the expert :D

6. Somehow the draw area gets blue for 1 sec when you draw one pixel on a Tablet - Touch & S-Pen.

off topic

1. console shows "waiting for templates to load .... piskel-boot.js:82:9
getPreventDefault() is deprecated you should use defaultPrevented. piskel-packaged-min-2015-11-05-03-08.js:2:0"

2. It would be nice to have more informations in the wiki "Development Environment". I use Ubuntu right now and I had to install more when the wiki told me + version problems. Maybe creating a install script?

3. I don't know how the js world works but I think they also have documentations so maybe adding those to the "Development Environment" will make things easier.

4. Using the browser zoom is messing up the draw position.



Tell me if you want me to open separate issues.

juliandescottes:
> if (typeof event.targetTouches && ... Moving the Mouse -> Error: TypeError: a.targetTouches is undefined

Actually I suggested using if ````(event.targetTouches && ...```` (notice there is no typeof). I can update it after merging.

> I don't know the difference between changedTouches and targetTouches

From the docs on MDN : targetTouches always returns the currently active touches, while changed touches has a behavior which depends on the event. Check https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent/changedTouches for more details.

> hasChangedTouchPoint - Maybe you add this on your own

Ok, I will after the merge !

> console shows "waiting for templates to load .... piskel-boot.js:82:9

I could get rid of that log, but I would say it doesn't bother anyone ?

> getPreventDefault() is deprecated you should use defaultPrevented. piskel-packaged-min-2015-11-05-03-08.js:2:0"

This warning actually comes from a lib (jquery). Will be gone when we can remove it :)

> Tell me if you want me to open separate issues.

Would be better, even if the answer is simple, it's easier to reference it if it comes back.

> It would be nice to have more informations in the wiki "Development Environment". I use Ubuntu right now and I had to install more when the wiki told me + version problems. Maybe creating a install script?

I am regularly doing new installs on different OS, and never had to do anything beyond what I mentioned in the wiki :/

Please open an issue explaining in details where you had trouble ? Was it with installing node ? grunt ? running the app.

> I don't know how the js world works but I think they also have documentations so maybe adding those to the "Development Environment" will make things easier.

Are you thinking about some kind of api documentation ? For the piskel's code ? For JS development in general ? Can you give me more details about the documentation you expected ? 

> Using the browser zoom is messing up the draw position

Can't reproduce on Firefox/Chrome at the moment, please open a separate issue for this.

juliandescottes:
And as a separate : I will merge this later today ! Thanks :)

PR #347, opened 2015-11-30
Blur disabled buttons

captainbrosset:
This fixes the issue on Firefox and it still works on Chrome. I haven't tested elsewhere.
Are there tests for this that I should change?


captainbrosset:
Not sure where those other commits come from ... I messed up my repo I think.

juliandescottes:
Thanks for the PR !

Sadly no, no tests are covering this :/ 

We only have : 
- unit tests using Karma+Jasmine testing utils, services (with limited coverage at the moment)
- drawing tests which test drawing-tools / frames / transforms (good coverage)

We are missing : 
- DOM interaction tests (à-la selenium)
- Controllers tests (mostly means the settings & dialogs controllers are not tested)

juliandescottes:
Merged in 6f5e2f13, closing.

PR #385, opened 2016-01-18
Introduces zooming towards/away from mouse

smiegrin:


juliandescottes:
Thanks for the PR! 
Sorry about the lack of feedback, I will have a look before the weekend.

smiegrin:
Is there anything that I need to adjust before this can be merged?

juliandescottes:
> Sorry about the lack of feedback, I will have a look before the weekend.

6 months later ... "Nice job, Julian" :/ 

Sorry again. So I did test your PR, and it does improve the behavior. But I still had one issue with it. It "feels" wrong if you start zooming while the canvas is zoomed out enough so that the lightgrey borders are displayed:

piskel-zoom

This is not directly related to your implementation here. Since the canvas is always centered when it is zoomed out, the relative position of my mouse cursor in the canvas will change between the moment I start zooming and the moment we can start offsetting the canvas. 

That being said this is the behaviour of other image editing software, so maybe it's OK to leave it as is.

src/js/controller/DrawingController.js
```diff
@@ -247,13 +247,37 @@
...
     this.setZoom_(this.renderer.getZoom() + step);
+    var newWidth = this.getContainerWidth_() / this.renderer.getZoom();
+    var newHeight = this.getContainerHeight_() / this.renderer.getZoom();
+    this.setOffset(
+      off.x + ((oldWidth - newWidth) * xRatio),
+      off.y + ((oldHeight - newHeight) * yRatio)
+    );
   };
 
   ns.DrawingController.prototype.decreaseZoom_ = function (zoomMultiplier) {
```
juliandescottes: Let's have a single method and remove duplication here: 

    ns.DrawingController.prototype.updateZoom_ = function (delta) {
        // your new before-code
        this.setZoom_(this.renderer.getZoom() + delta);
        // your new after-code
    }

And either use it in increase/decreaseZoom, or directly remove the increase/decreaseZoom methods and use updateZoom directly.

smiegrin:
Alrighty, I'll get to it. Thank you!

smiegrin:
Let me know if there's a better way in terms of style to put this together; I just kind of tweaked everywhere in the file to try and make the feature work.

juliandescottes:
The code change looks good! Merging. Thanks again for the PR and sorry about the review delay :)

PR #396, opened 2016-02-03
fix #392

### LINKED ISSUE #392: Settings drawers should close on ESCAPE key
The settings drawers are all the panels that open on the right of the application (application settings, resize, save, export, import).

Right now, the only way to close them is to click outside of the drawer.
It would be nice if hitting "ESCAPE" would do the same. This behavior is already implemented for dialogs.

### Hints

The controllers responsible for this area are in src/js/controller/settings. SettingsController.js is the main controller that handles the drawer. Individual controllers (for instance ApplicationSettingsController.js) are only responsible for the content displayed inside the drawer. They are loaded by the SettingsController.

Regarding keyboard shortcuts, you can have a look at src/js/controller/dialogs/DialogsController.js and how it uses "closePopupShortcut" :

    pskl.app.shortcutService.registerShortcut(
       this.closePopupShortcut, this.hideDialog.bind(this));

The available keyboard shortcuts are listed in src/js/service/keyboard/Shortcut.

### Testing

There is no framework to test this part of the application for now.

GMartigny:
fix #392 
Allow to close the options drawer with the escape key.

### DIFF
--- a/src/js/controller/settings/SettingsController.js
+++ b/src/js/controller/settings/SettingsController.js
@@ -33,6 +33,7 @@
 
   ns.SettingsController = function (piskelController) {
     this.piskelController = piskelController;
+    this.closeDrawerShortcut = pskl.service.keyboard.Shortcuts.MISC.CLOSE_POPUP;
     this.settingsContainer = document.querySelector('[data-pskl-controller=settings]');
     this.drawerContainer = document.getElementById('drawer-container');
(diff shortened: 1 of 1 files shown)


GMartigny:
Sorry I closed it too sooon

juliandescottes:
(was not sure why you closed it :) )

Thanks for looking into this. Looks good to me, merging!

PR #397, opened 2016-02-03
Issue #348 : colorpicker showing on top of tooltip

### LINKED ISSUE #348: Help tooltips should always be opened in the foreground
STR:
- open the color picker
- hover over the background color placeholder

Actual:
tooltips
The black tooltip that contains a help string appears behind the color picker

Expected:
It should appear at the top, always visible.

Related: I don't think the title attribute tooltip (the one that contains the current color name) should appear. Or if it needs too, it should be somewhere else.

GMartigny:
fix #348
Reorder z-indexes to display palette tooltip hover spectrum popover.
Remove title attribute on color selector and add a proposition as a replacement.

### DIFF
--- a/src/css/spectrum/spectrum.css
+++ b/src/css/spectrum/spectrum.css
@@ -13,7 +13,7 @@ License: MIT
     *display: inline;
     *zoom: 1;
     /* https://github.com/bgrins/spectrum/issues/40 */
-    z-index: 20000;
+    z-index: 9999994;
     overflow: hidden;
 }
 .sp-container.sp-flat {
--- a/src/css/spectrum/spectrum.css
+++ b/src/css/spectrum/spectrum.css
@@ -13,7 +13,7 @@ License: MIT
     *display: inline;
     *zoom: 1;
     /* https://github.com/bgrins/spectrum/issues/40 */
-    z-index: 9999994;
+    z-index: 20000;
     overflow: hidden;
 }
 .sp-container.sp-flat {
--- a/src/js/controller/PaletteController.js
+++ b/src/js/controller/PaletteController.js
@@ -1,7 +1,9 @@
 (function () {
   var ns = $.namespace('pskl.controller');
 
-  ns.PaletteController = function () {};
+  ns.PaletteController = function () {
+      this.displayer = $('.palette-value-displayer');
+  };
 
   /**
    * @public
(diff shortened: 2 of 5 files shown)


src/css/spectrum/spectrum.css
```diff
@@ -13,7 +13,7 @@ License: MIT
     *display: inline;
     *zoom: 1;
     /* https://github.com/bgrins/spectrum/issues/40 */
-    z-index: 9999994;
+    z-index: 20000;
```
juliandescottes: Was the previous value an issue ? 

If not I'd prefer not to diverge from the original spectrum css file.
If yes, you could add this custom style to the spectrum-overrides.css files. It contains all the "piskel" hacks.

src/js/controller/PaletteController.js
```diff
@@ -106,11 +110,21 @@
...
-    this.setTitleOnPicker_(color, colorPicker);
   };
 
-  ns.PaletteController.prototype.setTitleOnPicker_ = function (title, colorPicker) {
-    var spectrumInputSelector = '.sp-replacer';
-    colorPicker.next(spectrumInputSelector).attr('title', title);
+  /**
+   * @private
+   */
+  ns.PaletteController.prototype.showValueDisplayer_ = function (evt) {
```
juliandescottes: How about updating the content of the colopicker tooltip instead. It's already there so no extra UI needed. We're on a somewhat old version of the jquery tooltips, but updating is possible.

$(element).attr('data-original-title', newValue)
          .tooltip('fixTitle');

(tooltips are attached to the parent of "#color-picker")

src/css/spectrum/spectrum.css
GMartigny: This value was way to high for the Bootstrap tooltip (zindex: 30000).
Which one should be changed ?

src/js/controller/PaletteController.js
GMartigny: You mean changing the tooltip that say "Primary - left mouse button" ?

juliandescottes: Yes exactly. The tooltip could be on two lines, with the second line displaying the current color.

Sorry about the rushed comments/review. I had to leave in a hurry 5mn after starting the review.

src/css/spectrum/spectrum.css
juliandescottes: So if necessary let's move this to the override file.

I'm not sure we need the bootstrap tooltip to be displayed on top of the color picker. Do we have bootstrap tooltip inside the color picker ?

GMartigny: tooltip
Here's what we see at the moment. The informative tooltip for the red selector is cut in half.

It's neither practical nor important. There's a title "tooltip" on the palette handled by Spectrum.

src/js/controller/PaletteController.js
GMartigny: multi-line tooltip
I can't say I like it.

src/css/spectrum/spectrum.css
juliandescottes: Ok, so that's the tooltip from the color square "input/button" beneath the spectrum picker. IMO this tooltip should remain below the colorpicker.

However, it's true that the "current" color is not displayed anywhere here. Could be good for a follow up issue.

src/js/controller/PaletteController.js
juliandescottes: Sorry to read that :/ Anything in particular you dislike about this ?

The tooltip supports HTML so we could improve its layout. Maybe using the hex string would be nicer ?

GMartigny:
Set the bootstrap title on multi-line with a static part and the selected color.

juliandescottes:
Thanks for fixing this! Merging.

PR #399, opened 2016-02-04
Issue #350

### LINKED ISSUE #350: The bottom overflow image in the frame list is upside down
This happens for me on Windows with Firefox 44.

See this image:
frames

As you can see, the bottom of the frame list is supposed to have a nice overflow image that indicates that there are more frames below, but in my case it looks like it's upside down.

GMartigny:
fix #350
put background position x and y together for compatibility

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,9 @@ cache
 *.sublime-project
 *.sublime-workspace
 
+# netbeans project folder
+nbproject
+
 # git stackdumps
 *.stackdump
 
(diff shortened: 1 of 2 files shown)


juliandescottes:
Nice! Thanks a lot for fixing this.

Just a small comment : it would be nice if the first line of the commit message contained a bit of explanation. eg. "Fix #350 : fix CSS for framelist overflow images" or something like that.

Thanks for the many contributions anyway, that's great!

GMartigny:
Ok I'll be more descriptive :wink: 
And you're welcome, I have time to kill and I like your project.

Issue #349, opened 2015-11-30
The frame list shouldn't scroll sideways


captainbrosset:
This is a really minor one, but anyway.

Steps:
- create some frames
- start dragging a frame
- move it towards the right

On Firefox at least (haven't tested elsewhere): the frames list scrolls slightly to the left (even though there's no scrollbar). As a result, the left border of the frames isn't visible anymore.

![frames](https://cloud.githubusercontent.com/assets/1152698/11476827/1b572046-9785-11e5-8646-cbaacc1614f0.png)

juliandescottes:
Released in v0.7.0

PR #400, opened 2016-02-04
Issue #349 : Frame list slightly scroll on the right

GMartigny:
fix issue #349 and add options on the sortable widget
Frames can now only be dragged by the handle, not sure if wanted. (can be revert)

### DIFF
--- a/src/js/controller/FramesListController.js
+++ b/src/js/controller/FramesListController.js
@@ -128,8 +128,7 @@
       update: $.proxy(this.onUpdate_, this),
       items: '.preview-tile',
       axis: 'y',
-      tolerance: 'pointer',
-      handle: '.dnd-action'
+      tolerance: 'pointer'
     });
     $('#preview-list').disableSelection();
   };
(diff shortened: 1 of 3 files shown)


src/js/controller/FramesListController.js
```diff
@@ -124,9 +124,12 @@
   ns.FramesListController.prototype.initDragndropBehavior_ = function () {
 
     $('#preview-list').sortable({
-      placeholder: 'preview-tile-drop-proxy',
+      placeholder: 'preview-tile preview-tile-drop-proxy',
       update: $.proxy(this.onUpdate_, this),
-      items: '.preview-tile'
+      items: '.preview-tile',
+      axis: 'y',
```
juliandescottes: feels much nicer to use, thanks!

src/js/controller/FramesListController.js
```diff
@@ -124,9 +124,12 @@
...
+      tolerance: 'pointer',
+      handle: '.dnd-action'
```
juliandescottes: As you guessed, I'd prefer to keep the whole frame draggable.

juliandescottes:
Looks good!

Just keep the whole frame draggable and we'll merge this.

juliandescottes:
I didn't see the PR was updated ... sorry for the delay.
Thanks!

Issue #293, opened 2015-08-10
Export as PNG with X frames per row.


RainingChain:
Right now, you can only export as PNG with all frames side by side.

It would be great if one could specify the amount of frames per row. This could be done by adding a numeric input next to the 'Download PNG' button.

Here is a function that converts an image with all frames side by side to an image with the amount of rows specified.

```javascript
/*
  img:Image, image with all frames side by side
  totalFrame:Number, total amount of frames in the image
  framesPerRow:Number,
  return Image, image with the right amount of frames per row
*/
var convertImage = function(img,totalFrame,framesPerRow){
	if(framesPerRow === totalFrame)
		return img;
	
	var frameWidth = img.width/totalFrame;
	var frameHeight = img.height;
	var rowCount = Math.ceil(totalFrame/framesPerRow);
	
	var finalImgWidth = frameWidth * framesPerRow;
	var finalImgHeight = frameHeight * rowCount;
	
	var canvas = document.createElement('canvas');
	canvas.width = finalImgWidth;
	canvas.height = finalImgHeight;
	
	var ctx = canvas.getContext('2d');
	
	for(var i = 0; i < rowCount; i++){
		var cropWidth = finalImgWidth;
		if(i * finalImgWidth + finalImgWidth > img.width)	//incase row is not full (framesPerRow is not a factor of totalFrame)
			cropWidth = img.width - i * finalImgWidth;
		
		ctx.drawImage(img,
			i * finalImgWidth,0,
			cropWidth,frameHeight,
			0,i * frameHeight,
			cropWidth,frameHeight
		);
	}
	var newImg = new Image();
	newImg.src = canvas.toDataURL('image/png');
	return newImg;
}
```

bsmith555:
Please also consider an option for specifying the maximum pixel width of export file, or perhaps even both maximum dimensions.

LuciusSixPercent:
Maybe there could be an "advanced options" in the export dialog where you could specify rows, columns, dimensions, margins, padding, etc...

And I may be going too far here, but perhaps and option to export an XML with some metadata like frame duration and other custom tags. That could be a life changer for some! Although that may require changing how frames work internally and all... one can dream right?

juliandescottes:
Looks like there is a lot of interest for improvements to the import/export, so I will get started on that next.

@LuciusSixPercent : I like what you described with the XML, but is there some kind of standard widely recognized by game engines or should it be heavily customisable (XML, json etc ...)

[...]

PR #401, opened 2016-02-04
Issue #293 : Choose a number of frames per rows for the export

GMartigny:
First tackle on the issue #293
Add an input to select the number of frames per rows
Size the export output accordingly

This need to be reviewed


src/js/rendering/FramesheetRenderer.js
```diff
@@ -13,11 +13,13 @@
     }
   };
 
-  ns.FramesheetRenderer.prototype.renderAsCanvas = function () {
-    var canvas = this.createCanvas_();
+  ns.FramesheetRenderer.prototype.renderAsCanvas = function (framesPerRows) {
```
juliandescottes: PiskelRenderer and FramesheetRenderer are used by other callers than PngExport. 
framesPerRow should be defaulted to 0 if it's not provided.

(not related : I need to open another issue to get rid of "PiskelRenderer". Unnecessary and confusing ... congratulations for getting through this!)

juliandescottes: "should be defaulted to 0 if it's not provided" well actually it should default to this.frames.length (or Infinity)

src/js/controller/settings/exportimage/ImageExportController.js
```diff
@@ -17,6 +17,11 @@
     this.addEventListener(scalingFactorInput, 'input', this.onScalingFactorChange_);
     this.updateScalingFactorText_(scalingFactorInput.value);
 
+    var framesPerRowsInput = document.querySelector('.export-frames-per-rows-field');
+    framesPerRowsInput.value = pskl.UserSettings.get(pskl.UserSettings.FRAMES_PER_ROWS);
+    this.addEventListener(framesPerRowsInput, 'change', this.onFramesPerRowsChange_);
+    this.addEventListener(framesPerRowsInput, 'input', this.onFramesPerRowsChange_);
+
```
juliandescottes: I'm trying to think if there is a good reason to keep this in ImageExportController rather than in PngExportController ? Looking at the code I see we have the scaling factor logic in ImageExport as well, but I think this might be a leftover.

If you tried and it didn't work for some reason, just tell me what was the issue, so that we can file a follow up issue.

juliandescottes:
Thanks for tackling this and sorry for the late review! Have a look at the review comments (and don't hesitate to ping me by email if I forget to review, I seem to miss github notifications these days)

I have some ideas for tweaking the UI further, but we should discuss them in a follow up issue.
If @grosbouddha has some time to spare, could use your opinion here!

#### input number vs input text
<input number> has a bit of a clunky UI (can be styled in CSS, but Chrome only). I would prefer to use a plain <input text> with some JS validation, and support for up/down arrows to increment/decrement the value.

#### col x row
I would like to try a UI with two linked inputs, one for rows, one for columns. When exporting a 64 frames animation, the UI would show "Layout : [  1] row x [ 64] cols". Updating one of the field would change the other one so that we always show the user what the final grid will look like. This would remove the "0 to disable" as well => user would just have to enter "1" in the row input again.

src/js/controller/settings/exportimage/ImageExportController.js
GMartigny: I put it just next to the "scaling fator" controlling to follow the same logic. Should I move both on the export controller directly ?

juliandescottes: Just what you are adding for this feature, thanks!
I will file a cleanup issue for this controller afterwards.

juliandescottes:
#293 is now closed, so closing the PR here.

Issue #411, opened 2016-02-21
Export as ZIP: Pad frame numbers with leading zeroes


AndyBarron:
Lack of leading zeroes causes a naive string sort to do this:
`run_1, run_10, run_11, run_2, run_3, ...`
Certain animation tools rely on the frame names to be zero-padded like so:
`run_01, run_02, run_03, ... run_09, run_10, run_11, ...`

juliandescottes:
Thanks for filing! 
Seems like a good first bug, so I'll drop a few hints if anyone wants to take it.

The export to ZIP is currently handled in https://github.com/juliandescottes/piskel/blob/master/src/js/controller/settings/exportimage/PngExportController.js

We currently set the zip file names in two methods : mergedExport_ and splittedExport_.
While implementing this, the duplicated code between these two methods could be removed. 

The String pad method can be implemented in the PngExportController or in a /utils helper (either the existing core.js, or a new StringUtils.js file). A /utils helper would be a bit better, because we can easily unit test it. 

[For information, the whole zip export should be moved out of the controller into a service, and the Zip export should have its own controller. But no need to address that here.]

The padding should be relative to the number of frames. E.g : 
- if there are less than 10 frames, no padding
- more than 10 less than 100 : 01, 02, ..., 99.
- more than 100 less than 1000 : 001, 002, ... , 010, 011, ..., 999
- etc ... (don't think I ever saw a sprite made with piskel containing more than 1000 frames)

juliandescottes:
Released in v0.7.0

PR #433, opened 2016-03-23
Issue #411: Export as ZIP: Pad frame numbers with leading zeroes

mattdinthehouse:
I wrote up the code to address issue #411.

As suggested in the issue comments I added the padding function to a new StringUtils.js utility.

I don't know how you'd merge the mergedExport and splittedExport to remove the duplicate code (they're just a little bit too different) but I'm open to suggestions.


juliandescottes:
Thanks for fixing this! 
Works well and the code change looks good. 

> I don't know how you'd merge the mergedExport and splittedExport to remove the duplicate code 

Agreed :) Not sure what I had in mind!

Merging.

PR #440, opened 2016-04-06
C file export

clementleger:
This patch add a C file exporter. It allows to generate files with array of uint32_t representing the images. Embedding images in C programs can then be done easily.
The height of the drawer was modified to allow the C export button to fit.
Feel free to tell me if it needs reworks since I'm not used to javascript !

### DIFF
--- a/src/js/controller/settings/exportimage/CExportController.js
+++ b/src/js/controller/settings/exportimage/CExportController.js
@@ -1,0 +1,81 @@
+(function () {
+  var ns = $.namespace('pskl.controller.settings.exportimage');
+
+  var BLACK = '#000000';
+
+  ns.CExportController = function (piskelController) {
+    this.piskelController = piskelController;
+  };
+
+  pskl.utils.inherit(ns.CExportController, pskl.controller.settings.AbstractSettingController);
+
+  ns.CExportController.prototype.init = function () {
+
+    var downloadButton = document.querySelector('.c-download-button');
+    this.addEventListener(downloadButton, 'click', this.onCDownloadButtonClick_);
+  };
+
+  ns.CExportController.prototype.onCDownloadButtonClick_ = function (evt) {
+    var fileName = this.getPiskelName_() + '.c';
+    var cName = this.getPiskelName_().replace(" ","_");
+    var outputCanvas = this.getFramesheetAsCanvas();
+    var width = this.piskelController.getWidth();
+    var height = this.piskelController.getHeight();
+    
+    // Create a background canvas that will be filled with the transparent color before each render.
+    var background = pskl.utils.CanvasUtils.createCanvas(width, height);
+    var context = background.getContext('2d');
+    context.fillStyle = BLACK;
+
+    // Useful defines for C routines
+    var frameStr = '#include <stdint.h>\n\n';
+    frameStr += '#define '+ cName.toUpperCase() + '_FRAME_COUNT ' +  this.piskelController.getFrameCount() + '\n';
+    frameStr += '#define '+ cName.toUpperCase() + '_WIDTH ' + width + '\n';
+    frameStr += '#define '+ cName.toUpperCase() + '_HEIGHT ' + height + '\n\n';
+
+    frameStr += '/* Piskel \"' + this.getPiskelName_() + '\" */\n\n';
+
+    frameStr += 'uint32_t ' + this.getPiskelName_().toLowerCase();
+    frameStr += '_data[' + this.piskelController.getFrameCount() + '][' + width * height + '] = {\n';
+   
+    for (var i = 0 ; i < this.piskelController.getFrameCount() ; i++) {
+      var render = this.piskelController.renderFrameAt(i, true);
+      context.clearRect(0, 0, width, height);
+      context.fillRect(0, 0, width, height);
+      context.drawImage(render, 0, 0, width, height);
+      var imgd = context.getImageData(0, 0, width, height);
+      var pix = imgd.data;
+
+      frameStr += '{\n';
+      for (var i = 0; i < pix.length; i += 4) {
+	frameStr += this.rgbToCUint(pix[i], pix[i+1], pix[i+2]);
+	if (i != pix.length - 4)
+	  frameStr += ', ';
+	if (i % (width * 4) == 0)
+	  frameStr += '\n';
+      }
+      frameStr += '\n}';
+      
+    }
+    
+    frameStr += '};'
+    pskl.utils.BlobUtils.stringToBlob(frameStr, function(blob) {
+      pskl.utils.FileUtils.downloadAsFile(blob, fileName);
+    }.bind(this), 'application/text');
+ 
+  };
+
+  ns.CExportController.prototype.getPiskelName_ = function () {
+    return this.piskelController.getPiskel().getDescriptor().name;
+  };
+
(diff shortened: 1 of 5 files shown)


juliandescottes:
Thanks for the PR! 
The code change itself looks good, but I need to understand the export format a bit better before deciding how we merge this.

Is this somehow a standard for C games? It looks like a nice and convenient wrapper, but I'm wondering if it's actually used by the community?

We have to be careful with the options we present to users. The export panel is already a bit crowded, so if it's not a well established standard, we'll have to find another way of making this available. Could be hidden behind a preference, could have a popup for "custom/advanced" export formats etc...

clementleger:
Hello Julian, thanks for the comment. Regarding the export format, ithere is no "de facto" standard for C games (As far as I know) but this export format fits a lot of program. Indeed, the "piskel" is represented as an array of images (raw byte arrays). This is commonly used to directly embed sprites/images in embedded C program.
As you pointed out, this is not a common export format and as such, it should probably be located in a more advanced export panel. What would you recommend ?

clementleger:
Would you be ok having an "Advanced export" button at the bottom of the export panel which would be popping a dialog with various "less common" exporters ?

juliandescottes:
> Would you be ok having an "Advanced export" button at the bottom of the export panel which would be popping a dialog with various "less common" exporters ?

That's probably the way to go for now. A button or link that opens a dialog with a dropdown to select the export method => created issue #446 to handle this.
Let's review this PR now, no reason to block it !

(If #446 has not been addressed before the next release, I might hide the export format behind a pref)

src/js/controller/settings/exportimage/CExportController.js
```diff
@@ -0,0 +1,91 @@
...
+  ns.CExportController.prototype.init = function () {
+
+    var downloadButton = document.querySelector('.c-download-button');
+    this.addEventListener(downloadButton, 'click', this.onCDownloadButtonClick_);
+  };
+
+  ns.CExportController.prototype.onCDownloadButtonClick_ = function (evt) {
+    var fileName = this.getPiskelName_() + '.c';
+    var cName = this.getPiskelName_().replace(' ','_');
+    var outputCanvas = this.getFramesheetAsCanvas();
```
juliandescottes: Looks like outputCanvas is not used ?

src/js/controller/settings/exportimage/CExportController.js
```diff
@@ -0,0 +1,91 @@
...
+  ns.CExportController.prototype.rgbToCHex = function (r, g, b, a) {
+    var hexStr = '0x';
+    hexStr += ('00' + a.toString(16)).substr(-2);
+    hexStr += ('00' + b.toString(16)).substr(-2);
+    hexStr += ('00' + g.toString(16)).substr(-2);
+    hexStr += ('00' + r.toString(16)).substr(-2);
+    return hexStr;
+  };
+
+  ns.CExportController.prototype.getFramesheetAsCanvas = function () {
```
juliandescottes: Cf my comment about outputCanvas, can probably be removed.

src/js/controller/settings/exportimage/CExportController.js
```diff
@@ -0,0 +1,91 @@
...
+    frameStr += '#define ' + cName.toUpperCase() + '_FRAME_HEIGHT ' + height + '\n\n';
+
+    frameStr += '/* Piskel data for \"' + this.getPiskelName_() + '\" */\n\n';
+
+    frameStr += 'static const uint32_t ' + cName.toLowerCase();
+    frameStr += '_data[' + frameCount + '][' + width * height + '] = {\n';
+
+    for (var i = 0 ; i < frameCount ; i++) {
+      var render = this.piskelController.renderFrameAt(i, true);
+      context.clearRect(0, 0, width, height);
```
juliandescottes: I don't think clearRect is needed here if we do a fillRect with BLACK right after?

src/templates/settings/export.html
```diff
@@ -61,5 +61,12 @@
         <div class="gif-upload-status"></div>
       </div>
     </div>
+    <div class="settings-title">
+      Export as C File
+    </div>
+    <div class="settings-item">
+      <div class="settings-description">C file with frame rendered as array.</div>
+      <button type="button" class="button button-primary c-download-button">Download C file</button>
+    </div>
```
juliandescottes: I think we should temporarily move this section above the "GIF export" section. When using the "Get public URL" button of the GIF export, a preview of the image is inlined and makes the new section move up and down :)

src/js/controller/settings/exportimage/CExportController.js
```diff
@@ -0,0 +1,91 @@
...
+    var cName = this.getPiskelName_().replace(' ','_');
+    var outputCanvas = this.getFramesheetAsCanvas();
+    var width = this.piskelController.getWidth();
+    var height = this.piskelController.getHeight();
+    var frameCount = this.piskelController.getFrameCount();
+
+    // Create a background canvas that will be filled with the transparent color before each render.
+    var background = pskl.utils.CanvasUtils.createCanvas(width, height);
+    var context = background.getContext('2d');
+    context.fillStyle = BLACK;
```
juliandescottes: Is there any reason you fill the frame black before drawing? 
Why not keep the transparency ?

src/js/controller/settings/exportimage/CExportController.js
clementleger: Indeed !

src/js/controller/settings/exportimage/CExportController.js
clementleger: Yes, copy paste from GifExportController :/

src/js/controller/settings/exportimage/CExportController.js
clementleger: Yes, I took that from the GifExportController, but from your last comment, it appears not to be needed (both fillrect and clearrect).

src/templates/settings/export.html
clementleger: Noted ! I will move it.

src/js/controller/settings/exportimage/CExportController.js
clementleger: Actually, it was tied to the fact I did not use transparency in my app but it's more relevant to keep transparency for other apps.

src/js/controller/settings/exportimage/CExportController.js
juliandescottes: After reading this again, I think we can simplify it further: this.piskelController.renderFrameAt(i, true); actually returns a Canvas2d element, so you can directly call getContext('2d') to get the context etc ... No need to redraw it to a temporary canvas element! Sorry, lack of API documentation :/

[For information, if we kept the current approach we would still need at least the clearRect. The code is drawing every frame in the same canvas. Before drawing frame N+1 you need to cleanup the canvas from frame N. If frame N+1 has transparent pixels, they won't override the pixels from frame N.]

clementleger: Ok, don't worry for the lack of documentation. I modified it to remove the whole temporary canvas copy.

juliandescottes:
Thanks for updating ! (sorry for the delay, I probably missed a notification :/ )

clementleger:
Thanks Julian, I will try to do some mock up for "advanced export" feature next time :)

PR #445, opened 2016-04-09
Warp -> Wrap in move tool spelling

robmuh:
Obvious enough.


juliandescottes:
Ah, good catch thanks! 

Reading the label today, feels like it can still be improved. Don't hesitate if you have more suggestions!

Issue #375, opened 2015-12-27
Zoom where cursor is


staab:
Similar to the PC game Supreme Commander; I'd like to be able to zoom in on the edge of my drawing to do detail work, but zoom focuses on the center, and I can't find a way to pan the view.

juliandescottes:
To move the viewport you can either : 
- drag the white rectangle on the animated preview (top right)
- hold the middle mouse button (or wheel button) and move your cursor

staab:
Ah, that'll do the trick. Thanks.

juliandescottes:
Fixed with #385

PR #448, opened 2016-04-18
Issue #375 zoom at cursor

GMartigny:
Change the offset on zoom toward mouse position.

Two things to decides :

- Should it be only on zoom in ? (I feel it's better, but not sure)
- What should be the new offset ? (I choose average between current and mouse, could be closer to current position)


juliandescottes:
Feature already integrated. closing.

Issue #352, opened 2015-11-30
"selected color" marker is not visible if the color is yellow


captainbrosset:
Yes, I know, very nitpicking, but it occurred to me that if your palette has yellow in it, it will be very hard to see the "selected color" marker on it (the yellow triangle that appears in the lower left corner when you click on a color in the palette section).

Maybe some box-shadow could help? Maybe instead of a triangle over the color, a different border color around the color placeholder?

![palette-png](https://cloud.githubusercontent.com/assets/1152698/11477547/cae1130c-9788-11e5-959f-a476481cfa85.PNG)

juliandescottes:
Will be release in v0.8.0

juliandescottes:
![image](https://cloud.githubusercontent.com/assets/1141550/14726011/94eb1db2-081f-11e6-8e17-36efd68571c3.png)

PR #453, opened 2016-04-20
Issue #352 palette selected marker invisible

GMartigny:
A possible solution. 2 drawbacks :

- Can generate ugly color
- Unsupported by IE (fail silently)


juliandescottes:
Thanks for the proposal Guillaume! I'd prefer to keep a constant color for the selected color marker, though.

I tried something at https://github.com/juliandescottes/piskel/pull/456 . What do you think?

GMartigny:
No worry, it was just an dumb solution.

PR #485, opened 2016-06-15
Extract import logic to ImportService

islemaster:
Isolates import dialog logic from actual import logic, which makes headless import easier to implement.

This is a small subset of the changes I just made in https://github.com/code-dot-org/piskel/pull/4 that I think usefully separate concerns for the upstream as well.  Right now it's only used by the import dialog, but I don't think we're far from a world where we want the `FileDropperService` to delegate to something like this too.

### DIFF
--- a/src/js/service/ImportService.js
+++ b/src/js/service/ImportService.js
@@ -1,8 +1,8 @@
-/* Image and Animation import service supporting the import dialogs,
- * file dropper service, and headless import API. */
+/* @file Image and Animation import service supporting the import dialog. */
 (function () {
   var ns = $.namespace('pskl.service');
   /**
+   * Image an animation import service supporting the import dialog.
    * @param {!PiskelController} piskelController
    * @param {!PreviewController} previewController
    * @constructor
@@ -14,6 +14,8 @@
   };
 
   /**
+   * Given an image object and some options, create a new Piskel and open it
+   * for editing.
    * @param {!Image} image
    * @param {!Object} options
    * @param {!string} options.importType - 'single' if not spritesheet
--- a/src/js/service/ImportService.js
+++ b/src/js/service/ImportService.js
@@ -27,7 +27,7 @@
    * @param {function} [onComplete]
    */
   ns.ImportService.prototype.newPiskelFromImage = function (image, options, onComplete) {
-    onComplete = onComplete || function () {};
+    onComplete = onComplete || Constants.EMPTY_FUNCTION;
     var importType = options.importType;
     var frameSizeX = options.frameSizeX;
     var frameSizeY = options.frameSizeY;
--- a/src/js/service/ImportService.js
+++ b/src/js/service/ImportService.js
@@ -1,12 +1,13 @@
 /* Image and Animation import service supporting the import dialogs,
  * file dropper service, and headless import API. */
 (function () {
+  var ns = $.namespace('pskl.service');
   /**
    * @param {!PiskelController} piskelController
    * @param {!PreviewController} previewController
    * @constructor
    */
-  var ImportService = $.namespace('pskl.service').ImportService =
+  ns.ImportService =
       function (piskelController, previewController) {
     this.piskelController_ = piskelController;
     this.previewController_ = previewController;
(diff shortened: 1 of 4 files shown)


src/js/service/ImportService.js
```diff
@@ -0,0 +1,121 @@
+/* Image and Animation import service supporting the import dialogs,
+ * file dropper service, and headless import API. */
+(function () {
+  /**
+   * @param {!PiskelController} piskelController
+   * @param {!PreviewController} previewController
+   * @constructor
+   */
+  var ImportService = $.namespace('pskl.service').ImportService =
```
juliandescottes: For consistency I would keep using the var ns = $.namespace(...) as we do in other files. Unless of course there is a good reason to do it differently here.

Not against changing this convention but if we do, let's do it everywhere.

src/js/service/ImportService.js
```diff
@@ -0,0 +1,121 @@
+/* Image and Animation import service supporting the import dialogs,
+ * file dropper service, and headless import API. */
+(function () {
+  /**
+   * @param {!PiskelController} piskelController
```
juliandescottes: We are very bad at JSdoc for Piskel, and I'm all for improving this! 
I haven't seen the exclamation mark (!) pattern before. Where is this coming from, what does it mean?

src/js/service/ImportService.js
```diff
@@ -0,0 +1,121 @@
+/* Image and Animation import service supporting the import dialogs,
+ * file dropper service, and headless import API. */
```
juliandescottes: Not used by the file dropper, so better not mention it here.

src/js/service/ImportService.js
```diff
@@ -0,0 +1,121 @@
...
+   * @param {!string} options.importType - 'single' if not spritesheet
+   * @param {!number} options.frameSizeX
+   * @param {!number} options.frameSizeY
+   * @param {number} [options.frameOffsetX] only used in spritesheet imports.
+   * @param {number} [options.frameOffsetY] only used in spritesheet imports.
+   * @param {!boolean} options.smoothing
+   * @param {function} [onComplete]
+   */
+  ImportService.prototype.newPiskelFromImage = function (image, options, onComplete) {
+    onComplete = onComplete || function () {};
```
juliandescottes: nit: you could use Constants.EMPTY_FUNCTION here (really not pushing for it, it's just FYI)

juliandescottes:
Thanks a lot for the PR Brad, extracting the import logic to a service makes perfect sense!

I think I only have minor style nits. I'll spend more time on it later today though, so maybe wait before addressing them. Sorry for the half review, I wish Github had a way to have review drafts :(

src/js/service/ImportService.js
islemaster: Specifically, a "non-nullable type" [(usejsdoc.com)](http://usejsdoc.org/tags-type.html).

islemaster:
Thanks for the initial review; I'll start addressing style nits this afternoon, which shouldn't interfere too much with further review.  Some of the style discrepancies arose from my team objecting to those style decisions in our fork, but I'll happily push back and try to stay as consistent with upstream as possible so we can easily backport improvements.

src/js/service/ImportService.js
islemaster: Nice catch, my bad; during initial exploration I thought this service might be used by the file dropper too, but its import behavior is different enough that I didn't hook it up.  It might make sense in the future though, especially if `ImportService` expands to centralize import logic for .piskel files, palettes, etc.

src/js/service/ImportService.js
islemaster: As a more general comment, I've done sort of a lousy job with the JSDoc in this PR: I haven't written descriptions for the methods because several of them were moved wholesale and I'm not sure I understand their whole purpose very well.  That said, I find just annotating types helps me a lot when trying to understand new code.

src/js/service/ImportService.js
```diff
@@ -0,0 +1,124 @@
...
+        frameOffsetX,
+        frameOffsetY,
+        frameSizeX,
+        frameSizeY,
+        /*useHorizonalStrips=*/ true,
+        /*ignoreEmptyFrames=*/ true);
+  };
+
+  /**
+   * @param {canvas[]} images
```
juliandescottes: _images_ non-nullable?

juliandescottes:
Looks good to me, merging! Thanks for fixing the code style to match the rest of the codebase.

I don't have strong feelings regarding the current code style, we can change it so that it also feels comfortable on your side. We can discuss this in a separate issue if you'd like.

PR #488, opened 2016-06-19
Plane system

thejohncrafter:
Hello !
A friend of mine asked me to create a program that could edit pixel-art plane-based game assets.
So I came up with this : it is capable of editing three-dimensional assets with the same tools as Piskel, and is compatible with the original app.

Even if I know this is not the original purpose of the app, I thought that if you may be interested.
I think some people may like this feature, and that would be also cool for me because I wouldn't be the only one to maintain this project :smile: 

I know there remain some work to do in order to fully implement this into the website, but I wanted to know your opinion before going further.
I also know that you won't accept this PR immediately, but you can tell me what to do to make this fully compatible with the current website !

Hope you'll like this PR !


juliandescottes:
Thanks for the PR! 

I tried out your branch locally and it looks like it's working pretty well. Given the amount of changes that's a nice achievement :) 

First of all I need to say I know nothing about plane based assets. Is there any documentation or example I could check out to understand how this is used ? Same question about the plane offset feature.

My gut feeling is that if we merge this in Piskel, we will have to hide it behind a preference. One of the main strengths of Piskel is that it's easy to use for beginners & children. I would like to try to preserve that. 

The decision to merge or not will really depend on the additional maintenance. I need to review the code change in depths to get a better idea. 

(Really sorry to sound conservative, given the hard work you've put into this :/ )

thejohncrafter:
Good thing you liked this ;)

The plane-based system is something that I and my friend more or less invented, it's just a way of creating art objects : the idea is that we wanted to keep making pixel arts but we also wanted to add depth to our asset, and that's what we came up with.

We thought that would be a way to make 3d models without using editors like blender, because we want to create things as fast and simply as possible (and learning Blender would be too long :palm_tree: )

So no, I don't think there is some kind of documentation about plane-based assets, nor examples because we just did not start creating assets (and I don't think someone already did something like this)

And, to answer your question, the offset system is just a way to define the plane's position :wink:

I did not think that as a main feature for the editor, but as you said it could be a "hidden" preference, for people who may want to try !

I also know there still remains work to do in order to make this compatible with the whole project, but I'll try to do my best !

PR #508, opened 2016-07-21
Replace 'alt' buttons with 'option' buttons on Macs?

leereilly:
Embarrassingly, I had a hard time finding the <kbd>ALT</kbd> button on my Macbook. 

screen_shot_2016-07-20_at_6_12_00_pm

It might be easier to refer to is as the <kbd>OPTION</kbd> button for users on Macs? The standard keyboard looks like this:

image

This PR just replaces <kbd>ALT</kbd> button references with <kbd>OPTION</kbd> if the UserAgent looks like a Mac :metal:

### DIFF
--- a/src/js/controller/dialogs/CheatsheetController.js
+++ b/src/js/controller/dialogs/CheatsheetController.js
@@ -150,6 +150,7 @@
   ns.CheatsheetController.prototype.formatKey_ = function (key) {
     if (pskl.utils.UserAgent.isMac) {
       key = key.replace('ctrl', 'cmd');
+      key = key.replace('alt', 'option');
     }
     key = key.replace(/left/i, '&#65513;');
     key = key.replace(/up/i, '&#65514;');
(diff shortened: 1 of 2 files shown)


juliandescottes:
Thanks for the PR!

On my macbook (US layout), the keyboard layout is the same as on your picture.
But on my apple keyboard (US Intl. layout), the alt key only shows "alt" and ⌥ : 
image

From a quick search, it looks like "option" is mainly used on US keyboard layouts. I'll dig a bit more before deciding if we swap "alt" for "option" here :)

In OSX menus, it seems like keyboard shortcuts are always displayed using the symbols (⌘, ⌥ and ⇧ ). That could be another option but we'd need a font that displays them correctly.

juliandescottes:
Looks like most OSX software uses 'option' to document keyboard shortcuts, so I guess that's what makes sense for most users. Let's merge this.

Thanks for your contribution! 

By the way I love the pixelated octocat from the screenshot. Would nicely replace my current github logo on http://piskelapp.com if you're ok with sharing it :) 

image

leereilly:
I think that would look great! I wish I could take credit for the :octocat:, but it was created by @manaten - I'd double check with them first. They let me use it when [I asked last year](https://twitter.com/leereilly/status/668664689535008768), so chances are good 😺 

Image via http://blog.manaten.net/entry/github_pr_fetch:

687474703a2f2f692e696d6775722e636f6d2f436a34724d72532e676966

leereilly:
Lookes like @manaten gave you the thumbs up to use this image on http://piskelapp.com 👍 

image

PR #509, opened 2016-07-24
retina icons

juliandescottes:


juliandescottes:
Adding 2x versions of all png icons, as well as the SVG source files.

PR #517, opened 2016-08-06
Correct desktop window icon

epicabsol:
I found the logo at dest/prod/logo.png, while there was no
dest/logo.png. Fixes juliandescottes/piskel#515


juliandescottes:
Nice catch! Thanks for fixing this.

PR #521, opened 2016-08-09
Smooth scrolling for the layers list

epicabsol:
This smooths the scrolling introduced by juliandescottes/piskel@ba56a32, which finishes addressing juliandescottes/piskel#502.

The same idea could be applied to the frames list if we wanted to do that.


juliandescottes:
Thanks for the PR! 👍 

I'm not really keen on adding code in the LayersListController dedicated to smooth scrolling. I would prefer an isolated util.

ScrollBehavior has been added to the specs of scrollIntoView, and even though it is only implemented in Firefox for now, we could maybe look into using a polyfill here.
For instance https://github.com/iamdustan/smoothscroll ? No need to integrate this in the build, we could just add the library in src/lib. 

@epicabsol : What do you think?

juliandescottes:
Fixed by 71ec809a using  https://github.com/iamdustan/smoothscroll

PR #529, opened 2016-08-24
Change onPiskelReady to support multiple callback

david-szabo97:
`window.onPiskelReady` is now an array and functions can be pushed into it to be called when piskel is ready.

This should be merged in conjunction with https://github.com/juliandescottes/piskel-website/pull/9
Fix issue: https://github.com/juliandescottes/piskel/issues/505

### DIFF
--- a/src/piskel-boot.js
+++ b/src/piskel-boot.js
@@ -9,7 +9,11 @@
     version = '';
   }
 
-  window.onPiskelReady = function () {
+  if (!window.onPiskelReady) {
+    window.onPiskelReady = [];
+  }
+
+  window._onPiskelReady = function () {
     var loadingMask = document.getElementById('loading-mask');
     loadingMask.style.opacity = 0;
     window.setTimeout(function () {loadingMask.parentNode.removeChild(loadingMask);}, 600);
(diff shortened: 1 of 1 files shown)


src/piskel-boot.js
```diff
@@ -18,6 +22,11 @@
     delete window.pskl_exports;
     delete window.loadDebugScripts;
     delete window.done;
+
+    // Run Piskel ready callbacks
+    for (var i in window.onPiskelReady) {
```
juliandescottes: Do not use for ... in to loop on the array. either a regular for loop or a forEach.

src/piskel-boot.js
```diff
@@ -9,7 +9,11 @@
     version = '';
   }
 
-  window.onPiskelReady = function () {
+  if (!window.onPiskelReady) {
```
juliandescottes: nit: use a name that hints at an array? "piskelReadyCallbacks" or something like that.

src/piskel-boot.js
```diff
@@ -74,6 +83,6 @@
     }
 
     loadStyle('css/piskel-style-packaged' + version + '.css');
-    loadScript(script, 'onPiskelReady()');
+    loadScript(script, "_onPiskelReady()");
```
juliandescottes: nit: please use single quotes 
(just realized that files directly under src/ are not linted, will open another issue for that)

juliandescottes:
Thanks for the PR! Looks good overall just a few comments to address before merging

juliandescottes:
Thanks for the update, merging.

PR #537, opened 2016-08-31
New Piskel button

david-szabo97:
Can be integrated with the piskelapp.com with an extra init data parameter.
Will push a PR at the website repo to be in sync when this one gets merged.

Solving this issue: https://github.com/juliandescottes/piskel/issues/526

---

This change is [Reviewable](https://reviewable.io/reviews/juliandescottes/piskel/537)


juliandescottes:
Didn't try the PR yet, but I'm not a big fan of having a direct action coming from a button in the right sidebar. All buttons in this area should open the right panel/drawer. 

In earlier versions of Piskel, I had action buttons mixed in there.  It was a pain to use. You never remember which button will open the drawer, which button will trigger an action. And one of the two implies a bigger commitment from the user. After some time you find yourself always pausing before clicking to make sure you're not going to misclick and trigger some action. But if all buttons open the drawer, you don't care, if it was the wrong one you can just switch.

Long story short, if we want to integrate a "new piskel" button in the editor UI, it has to go somewhere else. It can still be on the right side, but it has to be clearly separated from the other buttons.

fafafrei:
I think the 'new piskel' button for the desktop app should be located where the 'create piskel' is located on the web app.  I agree that the sidebar probably is not the best place.  Is there any update on this, it seems to have been forgotten?

juliandescottes:
@fafafrei this feature was not necessarily about the desktop applications I think. But it's true having a "new piskel" button there would be nice.

I am not sure how it will work in the desktop app though. The current PR works using the assumption that creating a new piskel can be done by doing a window.open to a given link (which works fine for piskelapp.com where you can just window.open('http://piskelapp.com/p/create'). 

Looking at the nwjs documentation I guess we can use a similar approach with http://docs.nwjs.io/en/latest/References/Window/#windowopenurl-options-callback ? If someone wants to give it a shot and confirm this works we can repurpose this issue to add the new button in the desktop apps.

PR #545, opened 2016-09-12
Add a flag to hide layer controls from the UI

caleybrock:
Hi! As we are getting feedback on Gamelab (to be used in a middle school course for Code.org), we noticed that layers can be a challenging concept for younger students to grasp. We decided to remove it from the UI, but since this is more complicated than a simple css change, I made a flag that can be easily flipped on or off. With `ENABLE_MULTIPLE_LAYERS: true` there is no change to the current Piskel UI so I thought it could be useful in the original project. Thoughts?

Flag true (unchanged):
screen shot 2016-09-12 at 12 56 07 pm
Flag false:
screen shot 2016-09-12 at 12 56 35 pm

---
This change is [Reviewable](https://reviewable.io/reviews/juliandescottes/piskel/545)

### DIFF
--- a/src/js/controller/LayersListController.js
+++ b/src/js/controller/LayersListController.js
@@ -24,6 +24,15 @@
 
     $.subscribe(Events.PISKEL_RESET, this.renderLayerList_.bind(this));
     $.subscribe(Events.USER_SETTINGS_CHANGED, $.proxy(this.onUserSettingsChange_, this));
+
+    if (!Constants.ENABLE_MULTIPLE_LAYERS) {
+      this.rootEl.style.padding = '0px';
+      this.rootEl.style.visibility = 'hidden';
+      this.rootEl.style.height = '0px';
+      this.rootEl.style.width = '0px';
+      this.rootEl.style.margin = '0px';
+      this.rootEl.style.minHeight = '0px';
+    }
   };
 
   ns.LayersListController.prototype.renderLayerList_ = function () {
--- a/src/js/Constants.js
+++ b/src/js/Constants.js
@@ -52,5 +52,8 @@ var Constants = {
   // SERVICE URLS
   APPENGINE_SAVE_URL : 'save',
   IMAGE_SERVICE_UPLOAD_URL : 'http://piskel-imgstore-b.appspot.com/__/upload',
-  IMAGE_SERVICE_GET_URL : 'http://piskel-imgstore-b.appspot.com/img/'
+  IMAGE_SERVICE_GET_URL : 'http://piskel-imgstore-b.appspot.com/img/',
+
+  // Hide or show layers controls from the UI
+  ENABLE_MULTIPLE_LAYERS: true
 };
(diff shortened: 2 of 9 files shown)


src/js/controller/LayersListController.js
```diff
@@ -24,6 +24,15 @@
...
     $.subscribe(Events.USER_SETTINGS_CHANGED, $.proxy(this.onUserSettingsChange_, this));
+
+    if (!Constants.ENABLE_MULTIPLE_LAYERS) {
+      this.rootEl.style.padding = '0px';
+      this.rootEl.style.visibility = 'hidden';
+      this.rootEl.style.height = '0px';
+      this.rootEl.style.width = '0px';
+      this.rootEl.style.margin = '0px';
+      this.rootEl.style.minHeight = '0px';
+    }
```
juliandescottes: Why not `this.rootEl.style.display = 'none'; `

src/js/controller/LayersListController.js
```diff
@@ -24,6 +24,15 @@
 
     $.subscribe(Events.PISKEL_RESET, this.renderLayerList_.bind(this));
     $.subscribe(Events.USER_SETTINGS_CHANGED, $.proxy(this.onUserSettingsChange_, this));
+
+    if (!Constants.ENABLE_MULTIPLE_LAYERS) {
```
juliandescottes: I would move this check to the top of the function and bail out after hiding the element (no need to attach events for this controller if it doesn't get displayed:
```javascript
ns.LayersListController.prototype.init = function () {
  this.rootEl = document.querySelector('.layers-list-container');
  if (!Constants.ENABLE_MULTIPLE_LAYERS) {
    this.rootEl.style.display = 'none';
    return;
  }
  // ... attach events etc...
};

caleybrock:
This is good feedback, thanks! You brought up some good points that didn't come up because of how we are using Piskel in code.org. We are disabling shortcuts, and it's not possible for users of code.org to upload a piskel file (so they won't be able to load anything with layers). Since you're not sure about this even being useful in Piskel, I think we might take your feedback and just keep these changes simple and isolated on our fork. @islemaster thoughts?

islemaster:
@caleybrock what's your estimate of the time needed to handle all of these cases (Shortcuts.js, ApplicationSettingsController.js, and importing files with layer info)?  If it's a lot, let's go ahead and keep this change in our fork, making sure to annotate changed lines well.   Sorry for sending you down this road! While I like the idea of doing this in a configurable way, I can understand not wanting it upstream since it's an option piskelapp.com is unlikely to ever use.

islemaster:
Though this discussion does make me realize that we should _totally_ support importing .piskel files into Game Lab!

caleybrock:
Yeah I'd say lots of these suggestions are useful even if we don't merge them here. I'll make those on code-dot-org fork and circle back here if we feel it would be helpful to have upstream.

PR #560, opened 2016-10-29
Allows square brush size of up to 32 pixels using keyboard...

smiegrin:
…shortcuts. I guess the commit title was a little long.

I hope this is an acceptable intermediate step for larger brush sizes. I haven't looked into round brush shapes yet or how to integrate this better into the graphics of the piskel website with the buttons and all that.

Let me know if there's anything that needs to change to make this code compliant with the rest of the program's style!

---

This change is [Reviewable](https://reviewable.io/reviews/juliandescottes/piskel/560)

### DIFF
--- a/src/js/service/pensize/PenSizeService.js
+++ b/src/js/service/pensize/PenSizeService.js
@@ -2,7 +2,7 @@
   var ns = $.namespace('pskl.service.pensize');
 
   var MIN_PENSIZE = 1;
-  var MAX_PENSIZE = 1000000;
+  var MAX_PENSIZE = 32;
 
   /**
    * Service to retrieve and modify the current pen size.
--- a/src/js/service/pensize/PenSizeService.js
+++ b/src/js/service/pensize/PenSizeService.js
@@ -2,7 +2,7 @@
   var ns = $.namespace('pskl.service.pensize');
 
   var MIN_PENSIZE = 1;
-  var MAX_PENSIZE = 4;
+  var MAX_PENSIZE = 1000000;
 
   /**
    * Service to retrieve and modify the current pen size.
--- a/src/js/utils/PixelUtils.js
+++ b/src/js/utils/PixelUtils.js
@@ -76,33 +76,20 @@
      *
      * @param  {Number} row  x-coordinate of the original pixel
      * @param  {Number} col  y-coordinate of the original pixel
-     * @param  {Number} size >= 1 && <= 4
+     * @param  {Number} size >= 1 && <= 1000000
      * @return {Array}  array of arrays of 2 Numbers (eg. [[0,0], [0,1], [1,0], [1,1]])
      */
     resizePixel : function (col, row, size) {
-      if (size == 1) {
-        return [[col, row]];
-      } else if (size == 2) {
-        return [
-          [col, row], [col + 1, row],
-          [col, row + 1], [col + 1, row + 1]
-        ];
-      } else if (size == 3) {
-        return [
-          [col - 1, row - 1], [col, row - 1], [col + 1, row - 1],
-          [col - 1, row + 0], [col, row + 0], [col + 1, row + 0],
-          [col - 1, row + 1], [col, row + 1], [col + 1, row + 1],
-        ];
-      } else if (size == 4) {
-        return [
-          [col - 1, row - 1], [col, row - 1], [col + 1, row - 1], [col + 2, row - 1],
-          [col - 1, row + 0], [col, row + 0], [col + 1, row + 0], [col + 2, row + 0],
-          [col - 1, row + 1], [col, row + 1], [col + 1, row + 1], [col + 2, row + 1],
-          [col - 1, row + 2], [col, row + 2], [col + 1, row + 2], [col + 2, row + 2],
-        ];
-      } else {
-        console.error('Unsupported size : ' + size);
+      var pixels = [];
+      if(size > 1000000) size = 1000000;
+
+      for(j = 0; j < size; j++) {
+        for(i = 0; i < size; i++) {
+          pixels.push([col-Math.floor(size/2)+i,row-Math.floor(size/2)+j]);
+        }
       }
+
+      return pixels;
     },
 
     /**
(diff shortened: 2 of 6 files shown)


smiegrin:
It would seem, to the best of my knowledge, that one of the checks is limiting the size of the pen. Any guidance on what I can do would be welcome.

juliandescottes:
@smiegrin thanks for the PR, I'll take a look shortly. I've been without internet for 2 weeks, didn't mean to ignore this :)

src/js/service/pensize/PenSizeService.js
```diff
@@ -2,7 +2,7 @@
   var ns = $.namespace('pskl.service.pensize');
 
   var MIN_PENSIZE = 1;
-  var MAX_PENSIZE = 4;
+  var MAX_PENSIZE = 1000000;
```
juliandescottes: I'd prefer to keep the maximum to a reasonable value. 32 at most.

src/js/utils/PixelUtils.js
```diff
@@ -76,33 +76,23 @@
...
-          [col - 1, row + 1], [col, row + 1], [col + 1, row + 1], [col + 2, row + 1],
-          [col - 1, row + 2], [col, row + 2], [col + 1, row + 2], [col + 2, row + 2],
-        ];
-      } else {
-        console.error('Unsupported size : ' + size);
+      var pixels = [];
+      var i;
+      var j;
+
+      if (size > 1000000) { size = 1000000; }
```
juliandescottes: You shouldn't need to check for max pen size here as it is already handled in the pensizeservice.

src/js/utils/PixelUtils.js
```diff
@@ -76,33 +76,23 @@
...
+
+      for (j = 0; j < size; j++) {
+        for (i = 0; i < size; i++) {
+          pixels.push([col - Math.floor(size / 2) + i,row - Math.floor(size / 2) + j]);
```
juliandescottes: nit: add a space between "i," and "row"

src/js/service/pensize/PenSizeService.js
juliandescottes: (one additional comment about this: when the pen size gets really big, the rectangle/circle/line tools lag a lot - and the circle no longer resembles a circle. We should rework the tools so that they play nicely with bigger pen sizes.)

juliandescottes:
> It would seem, to the best of my knowledge, that one of the checks is limiting the size of the pen. 

So the PenSizeServiceTest is actively checking that setPenSize(5) keeps the pen size at 4. With your new implementation, you should update the test to check 1000001 (but as I mentioned, I'd like this limit to be seriously reduced :) ).
Anyway you get the idea, the test has to be updated to remain in sync with the current max pen size.

smiegrin:
I'll see what I can do! I have yet to dig around through the code that makes the interface, but I'll let you know if I have difficulty.

juliandescottes:
Thanks @smiegrin I was not sure if you wanted to take care of the UI part as well! Let's go for the simplest option here, which is to simply display the pen size in the biggest square (as shown in the screenshot I attached earlier)

A few pointers to get started here.

The file that is responsible for creating and updating the UI for the brush size widget is https://github.com/juliandescottes/piskel/blob/master/src/js/controller/PenSizeController.js. The CSS is in https://github.com/juliandescottes/piskel/blob/master/src/css/pensize.css.

One way you could tackle this is by setting an additional data-attribute on the last option (data-real-size ? data-big-size ?), if the pen-size is greater than 4. This attribute can be used in a ::before pseudo element with `content: attr(data-real-size);`

smiegrin:
Let me know if there's a better way to go about this feature; I had to do quite a bit of experimenting to find the labeling effect in the unfamiliar realm of css. In particular, I am not able to find how to center the number in its container or make the font more appealing while accommodating system-available fonts and such.
image

Thank you @juliandescottes for all the guidance! This app has been a blast to work with!

juliandescottes:
Thanks for the update! This is almost ready to merge.
For the styling let's add the following CSS rules to the `.pen-size-option::before` selector:

```css
.pen-size-option::before {
    /* ... previous rules ... */
    text-align: center;
    line-height: 12px;
    font-size: 90%;
}
```

We also need a bit more breathing space in order to display sizes up to 32. So let's increase the size of all the pen-size icons, by reducing the paddings by one:

```css
.pen-size-option[data-size="1"] {
    padding: 5px; /* ... was 6 before ... */ 
}
/* ... other sizes should be bumped as well ... */
.pen-size-option[data-size="4"] {
    padding: 2px; /* ... was 3 before ... */
}
``` 

On my machine this looks OK, let me know if that works for you!

test/js/service/pensize/PenSizeServiceTest.js
```diff
@@ -56,7 +56,7 @@ describe("PenSize test suite", function() {
 
     penSizeService.init();
     // MAX_VALUE is 4
```
juliandescottes: nit: update the comment here MAX_VALUE is 32

and update the name of the test "skips invalid value (outside of [1, 32])"

src/js/utils/PixelUtils.js
```diff
@@ -76,33 +76,21 @@
...
-          [col - 1, row + 1], [col, row + 1], [col + 1, row + 1], [col + 2, row + 1],
-          [col - 1, row + 2], [col, row + 2], [col + 1, row + 2], [col + 2, row + 2],
-        ];
-      } else {
-        console.error('Unsupported size : ' + size);
+      var pixels = [];
+      var i;
+      var j;
+
+      for (j = 0; j < size; j++) {
```
juliandescottes: nit: let's inline the declaration of i and j in the for loops here 
`for (var i = ....`

src/templates/drawing-tools.html
```diff
@@ -5,7 +5,7 @@
         <div class="pen-size-option" data-size="1"></div>
         <div class="pen-size-option" data-size="2"></div>
         <div class="pen-size-option" data-size="3"></div>
-        <div class="pen-size-option" data-size="4"></div>
+        <div class="pen-size-option" data-size="4" real-pen-size="6"></div>
```
juliandescottes: Any reason to add this in the default template?

juliandescottes:
Thanks for the follow-up @smiegrin ! Merging.

PR #567, opened 2016-11-22
Add a dropdown for preview size options [1x, best, full]

GMartigny:
[WIP] To start code review

https://github.com/juliandescottes/piskel/issues/362

---
This change is [Reviewable](https://reviewable.io/reviews/juliandescottes/piskel/567)


juliandescottes:
Thanks for the PR @GMartigny , this is a good starting point! Let's first discuss about the feature before diving into the code review.

1. Could we make the additional options appear horizontally? Right now, the tooltip is displayed over the options, which makes the feature difficult to use.

2. The "best" option should be updated when the sprite is resized (should be able to reuse onFrameSizeChange_ in the PreviewController as an entry point here). If the best option was selected for old-size, it should remain selected for new-size

Some additional points that would be nice to implement:

3. If the best option is 1x, it should be hidden to avoid duplication

4. If seamless mode is enabled, we should disable the widget and force the 1x preview size temporarily

5. If the frame size is more than 200 in height and/or width (ie can't fit in the preview at 1x), the widget should be disabled and we should force the "full" preview mode.

We can first work on 1&2 and deal with the rest later, up to you.

juliandescottes:
(By the way the travis CI job failed for your initial PR, but this was just a timeout of one of the DrawingTests. I retriggered the job and it was successful)

GMartigny:
> 1. Could we make the additional options appear horizontally? Right now, the tooltip is displayed over the options, which makes the feature difficult to use.

Drop-left (instead of drop-down) don't feels natural, so I choose to put tooltip to the left. I think it looks good, but I let decide.

> 2. The "best" option should be updated when the sprite is resized

done

> 3. If the best option is 1x, it should be hidden to avoid duplication
> 4. If seamless mode is enabled, we should disable the widget and force the 1x preview size temporarily
> 5. If the frame size is more than 200 in height and/or width (ie can't fit in the preview at 1x), the widget should be disabled and we should force the "full" preview mode.

I feel like the widget shouldn't be hidden to avoid confusion. So, I only disabled unfit options.

But leaving only one selected option isn't the best either. What's your opinion on this ?

6. I add that if best and full yeild the same result, we only keep "best" ratio as it hold the real zoom value.

juliandescottes:
Thanks for the update @GMartigny !

> I feel like the widget shouldn't be hidden to avoid confusion. So, I only disabled unfit options.
> But leaving only one selected option isn't the best either. What's your opinion on this ?

I agree, disabling is almost always better than hiding. I just was thinking about hiding a specific option in the widget. If best is also "x1", we end up showing x1 twice, I think it could also be confusing.

So as a summary I would say:
- disable the widget if seamless mode is on, or if the preview would not fit in 200x200
- hide the "best" option if it is redundant with x1
- hide the "full" option if it is redundant with "best" (as suggested in your last comment)

When the widget is disabled, it should still show the current zoom value, but nothing should happen when hovering it with the mouse. Ideally the tooltip should explain why it is disabled. If you manage to plug in this logic we can figure out which text would be the best later. It could also be slightly more greyed out, to show that it's disabled.

When the widget is enabled, it should at least have two valid options available.

How does that sound? 

> Drop-left (instead of drop-down) don't feels natural, so I choose to put tooltip to the left. I think it looks good, but I let decide.

I think this looks good. The only issue is that tooltips are sometimes displayed with an offset due to the animation. But that's acceptable I think.

Great job so far!

juliandescottes:
One last thing, it looks like your last commit makes the tests fail, on travis/node. Maybe you are using an API for which we need a new polyfill, but parsing quickly your changes I didn't see any.

I'll have another look tonight!

src/js/controller/preview/PreviewController.js
```diff
@@ -55,65 +66,100 @@
...
+  ns.PreviewController.prototype.updatePreviewSizeButtons_ = function () {
+    var fullZoom = this.calculateZoom_();
+    var seamlessModeEnabled = pskl.UserSettings.get(pskl.UserSettings.SEAMLESS_MODE);
+
+    if (seamlessModeEnabled) {
+      this.togglePreviewSizeButtonState_(this.previewSize.best, false);
+      this.togglePreviewSizeButtonState_(this.previewSize.full, false);
+      this.onChangePreviewSize_(this.previewSize.original.name);
+    } else {
+      if (Number.isInteger(fullZoom)) {
```
juliandescottes: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger 
Not available everywhere yet. Either drop the polyfill in utils/core.js or directly use `Math.floor(fullZoom) === fullZoom` here.

That's the reason why the tests are failing on Travis.

GMartigny: Damn it IE ...
Sorry, I lost the habits to check for compatibility.

GMartigny:
I've done a table to summarize all use-case because some are still unspecified.

Tell me if this match with what you had in mind. (and if not, feel free to edit)
https://docs.google.com/spreadsheets/d/17ELvPZ5zjZ32NVxmMwkFh1Gh_2sOf1m4qfWXEtoML1Q/edit?usp=sharing

(thanks for the kinds words)

juliandescottes:
Thanks for the summary. 

About the normal mode:
I think you're right to highlight the use case where the sprite is bigger than the preview size. The x1 zoom doesn't make much sense, forcing full (and thus disabling the widget) could be good here as well.

About the seamless mode:
For the seamless category, if we agree about disabling the widget, we don't need to worry about the available options right? When the image has a dimension > 200px it could be interesting to force "full" instead of "x1" (I don't think the seamless mode is super useful for big sprites anyway?). 

> Can't they be useful despite seamless mode active ?

They could to some extent, but I think the most important is to make sure that someone entering the seamless drawing mode can easily see the texture/pattern. If the preview/zoom remains set on best or full, the user might be confused about the feature? 

Related to that, maybe we should even force a zoom size dedicated to the seamless mode, that would be calculated so that we see at least 9 tiles (3x3). But that would be for another issue. For now I think enforcing x1 (or full is full < x1) is enough.

src/templates/preview.html
```diff
@@ -1,7 +1,12 @@
 <div id="animated-preview-container" class="preview-container">
   <div class="canvas-container-wrapper minimap-container">
     <div class="preview-contextual-actions">
-      <div class="preview-contextual-action  original-size-button" rel="tooltip" data-placement="bottom">1x</div>
+      <div class="preview-drop-down">
+        <div class="preview-disable-overlay" rel="tooltip" data-placement="left"></div>
+        <div class="preview-contextual-action  original-size-button" rel="tooltip" data-placement="left">1x</div>
+        <div class="preview-contextual-action  best-size-button" rel="tooltip" data-placement="left"></div>
+        <div class="preview-contextual-action  full-size-button" rel="tooltip" data-placement="left">Full</div>
```
GMartigny: Are you pro or cons dynamic adding in the DOM ? Thoses 3 can be appened at PreviewController's init.

juliandescottes: I don't mind either way, and the performance difference should not be noticeable here. I tend to leave it in the markup if I don't have to create similar elements dynamically. 

In this case I don't think you'll be creating elements on the fly, so I would probably just leave it in the markup. But if it's easier for you to generate it dynamically, go ahead.

juliandescottes:
@GMartigny ping :) This slipped off my radar, sorry! 
Did you want to work on it some more or should I start to review / amend? Thanks for working on this!

GMartigny:
I was sure that it was a closed case. But indeed, some more changes are needed.
You can however take a look and tell me if synthax bother you in some way.

juliandescottes:
Thanks for the update! Looks good, merging.

juliandescottes:
I'll just change the labels to something shorter :) Hope that's ok!

GMartigny:
That's your code man 😄 do as you wish.
Glad to add a new feature that I want to use.

PR #572, opened 2016-11-27
Circle Retool for Large Pen Sizes

smiegrin:
This vastly improves performance for the Circle/Ellipse tool when the pen size is particularly large. It isn't perfect yet -- the smaller ellipses behave a bit strange (~ 5 px), but it's a vast improvement from the lag pointed out on this thread: [https://github.com/juliandescottes/piskel/pull/560](url)

---
This change is [Reviewable](https://reviewable.io/reviews/juliandescottes/piskel/572)

### DIFF
--- a/src/css/pensize.css
+++ b/src/css/pensize.css
@@ -47,3 +47,7 @@
 .pen-size-option.selected {
   border-color: gold;
 }
+
+.pen-size-option.labeled:before {
+  content: attr(real-pen-size);
+}
--- a/src/js/controller/PenSizeController.js
+++ b/src/js/controller/PenSizeController.js
@@ -25,9 +25,17 @@
   };
 
   ns.PenSizeController.prototype.updateSelectedOption_ = function () {
+    pskl.utils.Dom.removeClass('labeled', this.container);
     pskl.utils.Dom.removeClass('selected', this.container);
     var size = pskl.app.penSizeService.getPenSize();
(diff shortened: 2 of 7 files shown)


juliandescottes:
Just tested out your branch and it is indeed much faster! Thanks for working on this.
Let's first merge the pull #560 and then we'll move on to reviewing / merging this one.

smiegrin:
Let me know if there's anything that needs to be added before this branch gets merged.

juliandescottes:
Thanks for the update. The circle tool is still producing circles with slightly thicker lines in some spots.

See for instance the difference between a 28x28 circle in the current release of piskel:
image

And one generated with this PR:
image

I think I would prefer us to stick with the current output, it feels cleaner. 
Your implementation seems to look better for all pen sizes > 1, so maybe we should keep the old implementation as a special case when pen size == 1 ?

juliandescottes:
On a sidenote, I'm about to close the release for v0.0.10. 
Would be nice to get this performance fix in before we wrap up :)

smiegrin:
I'm getting it together right now! =]

smiegrin:
I'm still working on it, but things aren't working as smoothly as I thought they would. I still plan on having it ready by tonight.

juliandescottes:
No worries, I probably won't get to the release before this weekend !

juliandescottes:
Thanks for updating the PR and fixing this @smiegrin 

Merging.

Issue #559, opened 2016-10-25
Layer name text wrapping disrupts layer list


epicabsol:
To reproduce, start a new piskel and create a new layer. Name it something long with spaces. Ensure that the top layer with the long name is selected. 
Observed behavior: The long layer name will wrap to the next line in the layer list, showing behind.
![screenshot 2016-10-25 at 2 33 39 pm](https://cloud.githubusercontent.com/assets/4462127/19705217/16e1ad4e-9ac0-11e6-8119-4caeb9d7a51c.png)
Additionally, if there is a word without spaces in the layer title, the layers list will grow horizontally.
![screenshot 2016-10-25 at 2 35 23 pm](https://cloud.githubusercontent.com/assets/4462127/19705296/8f88136e-9ac0-11e6-8014-cf8551984c97.png)

Expected behavior: The long layer name is truncated, perhaps with an ellipsis, or the layer list item grows vertically, but doesn't overwrite the next layer title.

juliandescottes:
Thanks for the report @epicabsol !

GMartigny:
@juliandescottes What's prefered : Elipsis or vertical grow ?

juliandescottes:
@GMartigny I think we should go for an ellipsis, and make sure that the title attribute is properly set with the name of the layer.

juliandescottes:
Fixed by @GMartigny

PR #577, opened 2016-12-01
Add ellipsis to overflowing layer item name #559

GMartigny:
If I understand well your last comment on the issue, you want a tooltip when the layer name is overflowing ?

It's going to be hard to have tooltip only in case of overflowing (not impossible though) and lost native CSS performance.

---
This change is [Reviewable](https://reviewable.io/reviews/juliandescottes/piskel/577)

### DIFF
--- a/src/css/toolbox-layers-list.css
+++ b/src/css/toolbox-layers-list.css
@@ -70,6 +70,8 @@
   padding: 0 0 0 10px;
   flex: 1 auto;
   white-space: nowrap;
+}
+.layer-item .layer-name.overflowing-name {
   overflow: hidden;
   text-overflow: ellipsis;
 }
@@ -79,7 +81,7 @@
 }
 
 .layer-item-opacity {
-  margin-right: 8px;
+  padding-right: 8px;
 }
 
 .current-layer-item,
--- a/src/css/toolbox-layers-list.css
+++ b/src/css/toolbox-layers-list.css
@@ -59,11 +59,19 @@
 
 .layer-item {
   position: relative;
+  display: flex;
   height:24px;
   line-height: 24px;
-  padding: 0 0 0 10px;
   border-top: 1px solid #444;
   cursor: pointer;
+}
+
+.layer-item .layer-name {
+  padding: 0 0 0 10px;
+  flex: 1 auto;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
 }
 
 .layer-item:hover {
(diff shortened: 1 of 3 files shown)


juliandescottes:
> If I understand well your last comment on the issue, you want a tooltip when the layer name is overflowing 

I don't think we need the usual jquery tooltip here. Simply set the title attribute on the layer element to have the default HTML tooltip displayed after some time. No need to make it depend on the overflow either.

src/templates/layers-list.html
```diff
@@ -36,7 +36,7 @@ <h3 class="toolbox-title layers-title">Layers
   <script type="text/template" id="layer-item-template">
     <li class="layer-item {{isselected:current-layer-item}}"
         data-layer-index="{{layerindex}}">
-        {{layername}}
+        <span class="layer-name">{{layername}}</span>
```
juliandescottes: Everything looks good! let's just add title="{{layername}}" here and it should be good to go :)

Thanks!

GMartigny:
Sorry for the spam :roll_eyes:, but it can be JQuery tooltip (init on the fly).

I see that it's never done in the code base. Is that on purpose or it just never had the need ?

juliandescottes:
I think we never had to dynamically add / remove tooltips yet. 

In this case I'm just leaning towards the default HTML tooltip, because I want it to be less intrusive than the jquery one. At least I would prefer if it was not displayed immediately (since it's a redundant information in case there is no ellipsis).

If you have a solution in mind using the jquery tooltip, feel free to update the PR, I'll test it and give you my feedback.

src/js/controller/LayersListController.js
```diff
@@ -99,6 +99,12 @@
     });
     var layerItem = pskl.utils.Template.createFromHTML(layerItemHtml);
     this.layersListEl.insertBefore(layerItem, this.layersListEl.firstChild);
+    if (layerItem.offsetWidth < layerItem.scrollWidth) {
+      $(layerItem).find('.layer-name')
+        .addClass('overflowing-name')
+        .attr('title', pskl.utils.TooltipFormatter.format(layer.getName()))
```
juliandescottes: I don't think you need the TooltipFormatter here, it is only used for complex tooltips.

src/css/toolbox-layers-list.css
```diff
@@ -70,6 +70,8 @@
   padding: 0 0 0 10px;
   flex: 1 auto;
   white-space: nowrap;
+}
+.layer-item .layer-name.overflowing-name {
```
juliandescottes: nit: let's add en empty line before this new selector.

juliandescottes:
Thanks, merging!

Issue #272, opened 2015-04-17
Feature: Copy from one piskel to another


dosjos:
Somethimes I create something in one piskel that I want to use in another piskel. To get the element over does seems impossible at the moment.

Idealy I would like to copy in one piskel and paste in another.

Possible workaround that don't work at the moment.
1. Create a piskel, draw something neat and save it.
2. Save it again under a different name
3. Edit the new save without destroying the first.

This would mean that when i rename a piskel, a new piskel is created, it is not actually renamed.

juliandescottes:
I see at least two different use cases here : 
- copy/paste between two sprites (so just copying a portion of a frame)
- import a complete sprite into another one

You confirm you're talking about the first option here ?

I think we can fix this quite easily. We just need to share the copy/paste content via LocalStorage and that should do it.

dosjos:
Yes, the first use case is what I have needed.
Although, someone might be happy for the second case as well

austinblanchard:
The second case you mentioned ("import a complete sprite into another one") is exactly what I came on here to find. Any implementation for this yet? Thanks!

juliandescottes:
@austinblanchard not yet, but thanks for the feedback. Helps to prioritize.

deweytyl:
Also looking for number 2, but seems like number 1 would solve the same situation.

lavadragon7i:
Is this thread still active? I kinda need #1

cbear1:
dude no 2 would be really helpful

juliandescottes:
> copy/paste between two sprites (so just copying a portion of a frame)

Was implemented by @PSeitz in #579 

> import a complete sprite into another one

Is being implemented in #285

clebs:
Hello,
sorry to reopen this, but I am trying to copy a selection from one piskel to another in the browser and it does not seem to work...
Was this feature removed or is incompatible with the latest version of chrome?

juliandescottes:
@clebs just tried on latest chrome on OSX and works for me.

My steps are:
- draw something in sprite 1
- select it
- copy (cmd/ctrl + C)
- switch to sprite 2
- paste (cmd/ctrl + V)

What are the exact steps you are doing?

clebs:
Hi @juliandescottes!

I investigated further and found out what is happening. It seems the target piskel canvas needs to be equal or bigger in size than the source canvas.

So if I try to copy a small selection of a big image into a small image (where the selection would fit) it does not work.
[...]

PR #579, opened 2016-12-04
Copy between piskels

PSeitz:
#272

---
This change is [Reviewable](https://reviewable.io/reviews/juliandescottes/piskel/579)

### DIFF
--- a/src/js/selection/SelectionManager.js
+++ b/src/js/selection/SelectionManager.js
@@ -95,7 +95,7 @@
     if (!this.currentSelection || !this.currentSelection.hasPastedContent) {
       if (window.localStorage.getItem('piskel.clipboard')) {
         this.currentSelection = JSON.parse(window.localStorage.getItem('piskel.clipboard'));
-      }else{
+      } else {
         return;
       }
     }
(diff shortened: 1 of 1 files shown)


juliandescottes:
Thanks a lot for your PR and sorry for the delay! 
That seems to work well and I think it's a good start for the feature. 

I can see potentially 3 improvements that we could tackle in follow up issues:
- allow the user to move the pasted selection after hitting ctrl+v
- local storage space is limited, so maybe try to store the selected pixels in a more compact format
- show a notification when a clipboard info is available from another sprite (we would need to be able to identify which clipboard data comes from which "session" but that should be doable)

Kuefet:
hello, so I am still stuck with one issue: I drew two different piskels, but can't copy and paste one from another. Are there any way to do that. note: I want to copy and paste only 1 frame. thanks a lot!

juliandescottes:
[below I refer to piskel 1 and piskel 2 to represent the piskel from which you want ot copy, and the target piskel]

If you just want to copy one frame (and a single layer), normally you can:
- in piskel 1, select the area you want to copy with the selection tool
- hit CTRL+C (CMD+C on OSX) to put the selection in your clipboard
- in piskel 2, hit CTRL+V (or CMD+V on OSX)

This should paste the selected content from piskel 1 into your current frame of piskel 2. You might want to create an additional frame or layer in piskel 2 before pasting, to be able to easily move/adjust the pasted content.

copy-paste-between-piskel

PSeitz:
I guess it's a litte weird to select something, by default it should probably copy the whole frame

Kuefet:
woaw.. thank you!

PR #580, opened 2016-12-04
Auto fix linting issues

PSeitz:
Auto fixing for minor issues like intendation is possbile. Though I guess, tests will be green, if not an automatic fix lint commit is created.

---
This change is [Reviewable](https://reviewable.io/reviews/juliandescottes/piskel/580)

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -73,6 +73,7 @@ module.exports = function(grunt) {
 
     jscs : {
       options : {
+        "fix": true,
         "config": ".jscsrc",
         "maximumLineLength": 120,
         "requireCamelCaseOrUpperCaseIdentifiers": "ignoreProperties",


juliandescottes:
I think it's great to make the "fix" option more easily available, but doing it by default will prevent the job on travis from catching the error. 

For instance I just played with this in https://travis-ci.org/juliandescottes/piskel/builds/183039433 . I added a commit that should fail jscs, but since jscs fixes the errors on the fly, the build doesn't fail and the bad commit goes unchecked.

What do you think about adding an additional grunt task "lint-fix" that would run all the linting tasks with autofix options when available?

PSeitz:
I think an additional task would be nice, it could also be automatically executed in an pre-commit hook

juliandescottes:
No longer using jscs, so closing this one. Thanks for the suggestion!

Issue #191, opened 2014-06-26
Import/export animation json data


devnewton:
It would  be nice to get json animation data with the zipped png images.

I use the following format for my own [animation software](http://devnewton.bci.im/en/softwares/nanim) and games:

https://github.com/devnewton/newton_adventure/blob/master/game/data/default_level_data/hero.json

hassanshaikley:
Bumping this : )

I am trying to use piskel with pixi js and JSON would be so great! : )

juliandescottes:
Just merged #581 ! Thanks.

PR #581, opened 2016-12-05
add pixi export

PSeitz:
#191

---
This change is [Reviewable](https://reviewable.io/reviews/juliandescottes/piskel/581)

### DIFF
--- a/src/js/controller/settings/exportimage/PngExportController.js
+++ b/src/js/controller/settings/exportimage/PngExportController.js
@@ -31,6 +31,7 @@
     this.columnsInput = document.querySelector('#png-export-columns');
 
     var downloadButton = document.querySelector('.png-download-button');
+    var downloadPixiButton = document.querySelector('.png-pixi-download-button');
     var dataUriButton = document.querySelector('.datauri-open-button');
 
     this.initLayoutSection_();
@@ -38,6 +39,7 @@
 
     this.addEventListener(this.columnsInput, 'input', this.onColumnsInput_);
     this.addEventListener(downloadButton, 'click', this.onDownloadClick_);
+    this.addEventListener(downloadPixiButton, 'click', this.onPixiDownloadClick_);
     this.addEventListener(dataUriButton, 'click', this.onDataUriClick_);
     $.subscribe(Events.EXPORT_SCALE_CHANGED, this.onScaleChanged_);
   };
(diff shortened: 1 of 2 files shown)


PSeitz:
This creates a zip with the spritesheet and a json metadatafile in the pixijs format.
But to download only the json metadatafile and separate it from the png spritesheet download may be better.

screen1

juliandescottes:
Thanks @PSeitz this looks great! 

I was about to suggest moving this to the "other" tab, but I guess it can be useful to set the number of columns for the export here, so let's leave it in the "PNG" tab. 

I think we should open another issue to rework a bit the UI of this tab. Maybe group the image download / data-uri options (or even remove the data-uri button completely ... this one got released by mistake, I wanted to have it only for development initially).

I am not really using Pixi personally, so if @hassanshaikley wants to also give some feedback that'd be great!

juliandescottes:
Code change looks good. Will merge soon.

src/js/controller/settings/exportimage/PngExportController.js
```diff
@@ -156,6 +158,52 @@
...
+        'spriteSourceSize': {'x': 0,'y': 0,'w': width,'h': height},
+        'sourceSize': {'w': width,'h': height}
+      };
+      frames[name + i + '.png'] = frame;
+    }
+
+    var json = {
+      'frames': frames,
+      'meta': {
+        'app': 'http://www.pisker.com',
```
juliandescottes: was this supposed to be piskelapp.com ? 

I guess we can just link to https://github.com/juliandescottes/piskel/

src/js/controller/settings/exportimage/PngExportController.js
```diff
@@ -156,6 +158,52 @@
...
+      var column = i % this.getColumns_();
+      var row = (i - column) / this.getColumns_();
+      var frame = {
+        'frame': {'x': width * column,'y': height * row,'w': width,'h': height},
+        'rotated': false,
+        'trimmed': false,
+        'spriteSourceSize': {'x': 0,'y': 0,'w': width,'h': height},
+        'sourceSize': {'w': width,'h': height}
+      };
+      frames[name + i + '.png'] = frame;
```
juliandescottes: I quickly tried this modifying the examples at https://github.com/pixijs/examples

The code was expecting frame numbers to be padded (ie piskel_name1.png -> piskel_name01.png). I'm not sure if this is something specific to the examples or that is more broadly expected for pixijs spritesheets?

hassanshaikley:
This is awesome! : )

juliandescottes:
Thanks for the feedback @hassanshaikley 

Merging and opening follow up issues to track my review comments.
Thanks for the great work @PSeitz !

src/js/controller/settings/exportimage/PngExportController.js
PSeitz: I don't know why they do this, in my example it works fine and the code is simpler. When the json+spritesheet is loaded, you usually reference the images by their names in pixijs.

src/js/controller/settings/exportimage/PngExportController.js
PSeitz: Yes, that should be fine

src/js/controller/settings/exportimage/PngExportController.js
juliandescottes: Yes, looks like this was really specific to this example. I think your implementation is fine.

Issue #563, opened 2016-11-18
Color preference. [Enhancement Request]


JFC263:
Hey I don't have any idea on how to use github but i love piskel and i would love if there would be a tiny feature in where you can choose in the settings which color type you rather use like RGB or #. Thats it.

Thank you for listening. @juliandescottes @grosbouddha

epicabsol:
I agree, although in the meantime, you can type 'rgb(r, g, b)' into the color entry textbox (or named colors, like 'sienna').

JFC263:
> I agree, although in the meantime, you can type 'rgb(r, g, b)' into the color entry textbox (or named colors, like 'sienna').<

Yes, but it still goes back to the other type of color and that cost effiency and time.

juliandescottes:
I had a quick look at what would be needed to suport this. I assume we are only talking about having changing the value displayed in the input of the colorpicker widget. This widget is a modified version of https://github.com/bgrins/spectrum . It supports a 'preferredFormat' configuration option. However, when setting this to 'hex', the transparent color becomes unselectable.

The spectrum widget used by piskel already has a substantial amounts of piskel specific modifications, so there is no easy upgrade path here. 

We could keep modifying the spectrum,js file though. The idea here would be to add an 'preferredInputFormat' configuration option to the widget and control this from an additional entry in the application settings panel.

The files to modify here would be:
- src/js/controller/settings/ApplicationSettingsController.js : add a select to pick the preferred color format
- src/js/utils/UserSettings.js : add a new user setting entry (e.g. COLOR_FORMAT)
- src/js/controller/PaletteController.js : listen to user settings changes (see other files listening to Events.USER_SETTINGS_CHANGED as a reference)
- src/js/lib/spectrum/spectrum.js (see https://github.com/juliandescottes/piskel/blob/master/src/js/lib/spectrum/spectrum.js#L674)

I can go into more details if anyone wants to pick this up.

JFC263:
>I had a quick look at what would be needed to suport this. I assume we are only talking about having changing the value displayed in the input of the colorpicker widget. 

Yes actually is just that.

PR #599, opened 2016-12-28
Add a setting for changing the color format shown in the color picker

zoeesilcock:
This is a WIP pull request for the following issue: #563 

From what I can tell this covers the feature request as needed without breaking anything but I do want some input on it since this is my first attempt at contributing to this project.

Dropdown in the settings:
settings

The color picker with decimal color format:
picker

I was initially listening for changes to the user settings in `PaletteController.js` as suggested in the issue. My guess was that the idea would be to update the pickers when the setting changes so that the UI stays up to date. It turned out to not be necessary for that reason, perhaps there was some other reason for listening to events there?

---
This change is [Reviewable](https://reviewable.io/reviews/juliandescottes/piskel/599)

### DIFF
--- a/src/js/utils/UserSettings.js
+++ b/src/js/utils/UserSettings.js
@@ -17,6 +17,7 @@
     EXPORT_TAB: 'EXPORT_TAB',
     PEN_SIZE : 'PEN_SIZE',
     RESIZE_SETTINGS: 'RESIZE_SETTINGS',
+    COLOR_FORMAT: 'COLOR_FORMAT',
     KEY_TO_DEFAULT_VALUE_MAP_ : {
       'GRID_WIDTH' : 0,
       'MAX_FPS' : 24,
@@ -39,7 +40,8 @@
         maintainRatio : true,
         resizeContent : false,
         origin : 'TOPLEFT'
-      }
+      },
+      COLOR_FORMAT: 'hex',
     },
 
     /**
--- a/src/templates/settings/application.html
+++ b/src/templates/settings/application.html
@@ -58,6 +58,14 @@
         <input type="text" class="textfield  textfield-small  max-fps-input" autocomplete="off" name="max-fps"/>
       </div>
 
+      <div class="settings-item">
+        <label for="color-format">Color format</label>
+        <select id="color-format" class="color-format-select">
+          <option value="hex">Hex</option>
+          <option value="rgb">RGB</option>
+        </select>
+      </div>
+
       <input type="submit" class="button button-primary" value="Apply settings" />
     </div>
   </form>
--- a/src/css/settings-application.css
+++ b/src/css/settings-application.css
@@ -54,7 +54,8 @@
   text-align: center;
 }
 
-.grid-width-select {
+.grid-width-select,
+.color-format-select {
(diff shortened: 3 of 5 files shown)


src/templates/settings/application.html
```diff
@@ -58,6 +58,14 @@
         <input type="text" class="textfield  textfield-small  max-fps-input" autocomplete="off" name="max-fps"/>
       </div>
 
+      <div class="settings-item">
+        <label for="color-format">Color format</label>
+        <select id="color-format" class="color-format-select">
+          <option value="0">Hex</option>
+          <option value="1">Decimal</option>
```
juliandescottes: I think `rgb(r,g,b)` is usually referred to as the RGB model or format

src/templates/settings/application.html
```diff
@@ -58,6 +58,14 @@
         <input type="text" class="textfield  textfield-small  max-fps-input" autocomplete="off" name="max-fps"/>
       </div>
 
+      <div class="settings-item">
+        <label for="color-format">Color format</label>
+        <select id="color-format" class="color-format-select">
```
juliandescottes: Small UI nit: in settings-application.css, there is a CSS rule applied to `grid-width-select`. 
We want the same rule to be applied to this element for consistency.

https://github.com/juliandescottes/piskel/blob/master/src/css/settings-application.css#L57-L59

src/js/utils/UserSettings.js
```diff
@@ -39,7 +40,8 @@
         maintainRatio : true,
         resizeContent : false,
         origin : 'TOPLEFT'
-      }
+      },
+      COLOR_FORMAT: 0,
```
juliandescottes: I'm being super nitpicky, but I would prefer if the preference was a readable key 'hex', 'rgb', and  if the constant was a map:

```javascript
COLOR_FORMATS: {
  'hex' : 'hex',
  'rgb' : 'rgb'
}
```

zoeesilcock: I used an array for easy converting from the 0 based index that you get from the select input. I was trying to mimic how the GRID_WIDTH setting works, but I couldn't find the place that it's index is changed into something more representative.

I'll dig into this one tomorrow and push once I'm done.

juliandescottes: I see, you're right this is how the grid select is working.

You don't have to use integer values in the input, you can use value="hex" and value="rgb".
Then you will directly get this string as evt.target.value.

This should do the trick but I haven't tested :) Let me know if you have any trouble with this, will be happy to help!

zoeesilcock: Oh yeah of course, silly me. That way I can skip the constant entirely. Thanks, I'll fix it tomorrow.

src/templates/settings/application.html
zoeesilcock: I added the `.color-format-select` selector to the CSS file.

juliandescottes:
Ah sorry I missed your last update, will merge later today !

juliandescottes:
Thanks @zoeesilcock , will be released with 0.10.0 !

Issue #363, opened 2015-12-08
After moving a frame using drag and drop, the wrong frame is selected (Firefox)


juliandescottes:
Steps to reproduce : 
- open Piskel in Firefox
- open a piskel with 4 distinct frame
- using the mouse/drag-and-drop, move the first frame to the another position

Expected : the moved frame should be selected
Actual : the new 'first' frame gets selected.

Some investigation : 
We have a moved frame F, going from index i1 to i2.
After the mouseup releasing F, the currentFrame is set to the new index of F, which is i2 At the next call of the render loop, all UIs will redraw accordingly.
However on Firefox a click event is also fired after the mouseup event, which triggers another call to setCurrentFrame. This one is based on the data-attribute of the event target. Since the UI has most likely not had time to redraw yet, F's data-attribute is still i1.

Could be a bug/limitation of jquery's sortable util.

juliandescottes:
Fixed by @zoeesilcock

PR #600, opened 2016-12-29
Workaround for incorrect frame selection after re-ordering on Firefox

zoeesilcock:
This commit is related to the following issue: #363

When a frame is dropped both `mouseup` and `click` events are triggered on Firefox. This is somwhat expected behaviour but other browsers do not trigger the `click` event.

One would expect jquery-ui to work the same across all browsers but when they got a bug report they decided to not change anything:
https://bugzilla.mozilla.org/show_bug.cgi?id=787944

The most comon workaround appears to be to use the `clone` helper in jquery-ui sortable. Unfortunately this doesn't work because the cloned frame doesn't keep the contents of the canvas. This seemed like the cleanest workaround, here are a few others:
http://stackoverflow.com/questions/947195/jquery-ui-sortable-how-can-i-cancel-the-click-event-on-an-item-thats-dragged

---
This change is [Reviewable](https://reviewable.io/reviews/juliandescottes/piskel/600)

### DIFF
--- a/src/js/controller/FramesListController.js
+++ b/src/js/controller/FramesListController.js
@@ -15,6 +15,7 @@
 
     this.redrawFlag = true;
     this.regenerateDomFlag = true;
+    this.justDropped = false;
 
     this.cachedFrameProcessor = new pskl.model.frame.CachedFrameProcessor();
     this.cachedFrameProcessor.setFrameProcessor(this.frameToPreviewCanvas_.bind(this));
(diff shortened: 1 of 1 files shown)


src/js/controller/FramesListController.js
```diff
@@ -213,6 +215,17 @@
 
   /**
    * @private
+   */
+  ns.FramesListController.prototype.onSortableStop_ = function (event, ui) {
+    this.justDropped = true;
+
+    this.resizeTimer = window.setTimeout($.proxy(function() {
```
juliandescottes: We don't reuse/clear this timer at any point so no need to store it. And resizeTimer is not really appropriate here :)

zoeesilcock:
That's an odd typo that I missed. I'm guessing that I copy/pasted a timer from elsewhere in your code and forgot to remove the variable. Could you remove it directly in master or should I make a new pull request?

juliandescottes:
Yup, already merged your commit and fixed the typo.

zoeesilcock:
Oh great, thanks!

Issue #601, opened 2016-12-29
Pressing "d" for default colors doesn't fully work.


TALANTO:
Hello I want to thank you for the great app and share few bugs I've found to make it even better!

When I change the left and right mouse button colors and I press "d" for the default colors they are actually changed but it's not shown in the bottom left corner:

![screenshot from 2016-12-29 13-20-40](https://cloud.githubusercontent.com/assets/16134110/21544825/69526130-cdca-11e6-9df0-251951d40bff.png)

juliandescottes:
Thanks for the bug report! 

The issue comes from PaletteController.js and the resetColors method. An option is to remove the reset method from the SelectedColorsService (since it doesn't fire the expected events anyway) and implement PaletteController::resetColors in the same way as PaletteController::swapColors

juliandescottes:
Fixed by #605 , thanks to @Sal0hcine

PR #605, opened 2017-01-05
Pressing "d" for default colors doesn't fully work. #601

Sal0hcine:
Removed reset from SelectedColorsService and implemented the reset in PaletteController

---
This change is [Reviewable](https://reviewable.io/reviews/juliandescottes/piskel/605)


juliandescottes:
Looks great @Sal0hcine ! 
Thanks a lot for fixing this.

Issue #285, opened 2015-06-05
Import multiple files as individual files.


Lazik:
Hi, I'd like to be able to import multiple files as individual frames in Piskel.
I get an animation spritesheet from my friends and then I have a program that automatically cut them as frame, I can save it where each frame is a file_xxx. I'd like to be able to load each file as a different frame.

Love the program, 
thanks

juliandescottes:
Thanks for the suggestion. I assume we are only talking about the image import here.

What is the format of the spritesheet you get ? If it's an image, maybe you could use the feature developed by @JALissiak ( see comments at https://github.com/juliandescottes/piskel/pull/284) ? I will update piskelapp.com with this feature later today.

That being said, there are two things we can do here : 

- import multiple files at once : you could select several files when doing the import, which would then be imported as different frames. The images would need to all have the same dimensions, and the frame order would be based on the alphabetical order of the filenames I guess.

- append imported image to the current sprite : import an image as a new frame. If the size doesn't match, image would be cropped. In case the sprite as several layers, the imported image would be imported in one layer, leaving the others empty

JFC263:
Thank you for referencing this issue when you closed mine and by what I am seeing a LOT of people are needing this feature so I understand if you need to constantly close more and more of the same problem are being created so fast and almost all of them are the same clone.


                                                                                                        Thanks for the support of your community.

                                                                                                                                    @JFC263

PewPewPerson:
So... I still don't understand how the merging of two piskels works.

juliandescottes:
@PewPewPerson this feature is not implemented yet.

EricEisaman:
I would like this feature in order to create a sprite atlas containing all the sprites required by a game. I would be able to pick and choose a collection of community contributed sprites to be consolidated into one atlas.  I will significantly reduce load times if I can load a single atlas on a worker thread in my application.

PR #620, opened 2017-01-22
Allow adding multiple frames with drag and drop

zoeesilcock:
This is a partial solution to issue number #285.

When dropping multiple images it creates a new frame for each one. It works great when all images are the same size and within the size of the piskel document. When the images aren't the size of the document they are positioned in the same way as single image import.

It might be nice to allow multiple image import using the import button but that would be a separate feature. This would be a big time saver for myself, please try it out and let mew know if I should change anything.

---
This change is [Reviewable](https://reviewable.io/reviews/juliandescottes/piskel/620)

### DIFF
--- a/src/js/service/FileDropperService.js
+++ b/src/js/service/FileDropperService.js
@@ -27,6 +27,8 @@
     };
 
     var files = event.dataTransfer.files;
+    this.isMultipleFiles_ = (files.length > 1);
+
     for (var i = 0; i < files.length ; i++) {
       var file = files[i];
       var isImage = file.type.indexOf('image') === 0;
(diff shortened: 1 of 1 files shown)


juliandescottes:
That looks good! 

Makes me think we should make the drag and drop more discoverable. Maybe we should have a bit of text in "Import" panel explaining that if you want to insert images in your current animation, you can simply dndrop them.

FYI I'm working on a new version of the import, that will allow you to merge your imports with your current sprite (taking care of resizing if needed). When it's ready we could plug it with the FileDropperService.

I'll test quickly but the code change looks good.

zoeesilcock:
Sounds good! Our workflow includes several tools and these type of features will make it easier for us to move our images between Piskel and other tools.

juliandescottes:
Works great! Thanks @zoeesilcock

PR #633, opened 2017-02-03
Rename layer with dblClick

GMartigny:
This may need some rework, so as always, leave you comment on this.

---
This change is [Reviewable](https://reviewable.io/reviews/juliandescottes/piskel/633)

### DIFF
--- a/src/js/controller/LayersListController.js
+++ b/src/js/controller/LayersListController.js
@@ -1,20 +1,20 @@
 (function () {
   var ns = $.namespace('pskl.controller');
 
-  var TOGGLE_LAYER_SHORTCUT = 'alt+L';
-
   ns.LayersListController = function (piskelController) {
     this.piskelController = piskelController;
-    this.layerPreviewShortcut = pskl.service.keyboard.Shortcuts.MISC.LAYER_PREVIEW  ;
+    this.layerPreviewShortcut = pskl.service.keyboard.Shortcuts.MISC.LAYER_PREVIEW;
   };
 
   ns.LayersListController.prototype.init = function () {
     this.layerItemTemplate_ = pskl.utils.Template.get('layer-item-template');
     this.rootEl = document.querySelector('.layers-list-container');
     this.layersListEl = document.querySelector('.layers-list');
+    this.currentLayerEl = null;
     this.toggleLayerPreviewEl = document.querySelector('.layers-toggle-preview');
 
     this.rootEl.addEventListener('click', this.onClick_.bind(this));
+    this.rootEl.addEventListener('dblclick', this.onDoubleClick_.bind(this));
     this.toggleLayerPreviewEl.addEventListener('click', this.toggleLayerPreview_.bind(this));
 
     this.initToggleLayerPreview_();
@@ -50,7 +50,6 @@
 
   ns.LayersListController.prototype.updateButtonStatus_ = function () {
     var layers = this.piskelController.getLayers();
-    var currentLayer = this.piskelController.getCurrentLayer();
     var index = this.piskelController.getCurrentLayerIndex();
 
     var isLast = index === 0;
@@ -105,6 +104,9 @@
         .attr('title', layer.getName())
         .tooltip();
     }
+    if (isSelected) {
+      this.currentLayerEl = layerItem;
+    }
   };
 
   ns.LayersListController.prototype.onClick_ = function (evt) {
@@ -113,8 +115,7 @@
     if (el.classList.contains('button')) {
       this.onButtonClick_(el);
     } else if (el.classList.contains('layer-name')) {
-      index = pskl.utils.Dom.getData(el, 'layerIndex');
-      this.piskelController.setCurrentLayerIndex(parseInt(index, 10));
+      this.focusLayerElement_(el.parentNode);
     } else if (el.classList.contains('layer-item-opacity')) {
       index = pskl.utils.Dom.getData(el, 'layerIndex');
       var layer = this.piskelController.getLayerAt(parseInt(index, 10));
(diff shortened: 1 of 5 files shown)


src/js/controller/LayersListController.js
```diff
@@ -123,14 +124,50 @@
...
+      return false;
+    }, this);
+
+    var input = this.currentLayerEl.querySelector('.rename-layer-input-text');
+    input.value = layer.name;
+    pskl.utils.Event.addEventListener(input, 'blur', this.renderLayerList_, this);
+    pskl.utils.Event.addEventListener(input, 'keydown', function(evt) {
+      var eventKey = pskl.service.keyboard.KeyUtils.createKeyFromEvent(evt);
+
+      if (eventKey.key === 'ESC') {
```
GMartigny: I felt the need to close form when ESC key is pressed, but key events are dismiss if fire inside an input.
I'm not cool with this solution, but I don't know if there's any.

src/js/controller/LayersListController.js
```diff
@@ -152,8 +189,10 @@
     } else if (action == 'merge') {
       this.mergeDownCurrentLayer_();
     } else if (action == 'edit') {
-      this.renameCurrentLayer_();
+      this.openCurrentLayerRenameForm_();
+      return;
```
GMartigny: This can be confusing to use return as function "stopper", but you do it too so I guess you're ok with that.

src/js/controller/LayersListController.js
```diff
@@ -152,8 +189,10 @@
...
     }
+    this.renderLayerList_();
```
juliandescottes: Why is this needed? 
- moveLayerUp triggers a reset
- moveLayerDown triggers a reset
- createKayer triggers a reset
- removeCurrentLayer triggers a reset
- mergeDown... triggers a reset (and also calls renderLayersList manually which is not needed)

src/js/controller/LayersListController.js
```diff
@@ -113,8 +115,7 @@
     if (el.classList.contains('button')) {
       this.onButtonClick_(el);
     } else if (el.classList.contains('layer-name')) {
-      index = pskl.utils.Dom.getData(el, 'layerIndex');
-      this.piskelController.setCurrentLayerIndex(parseInt(index, 10));
+      this.focusLayerElement_(el.parentNode);
```
juliandescottes: I guess you had conflicts between the click and doubleClick handlers, which is why you had to remove the RESET event when setting the layer index.

What was the issue exactly ? Was the second click event fired after the dblclick event, causing the markup to be regenerated ?

GMartigny:
You're right about the features, I was looking to add dblClick layer name editing and inline editing. But for the scrolling weirdness, it's a happy coincidence.

The biggest problem was that the whole DOM list was rebuild at every click (selecting layer). So, dblClick events couldn't be fired as it's not the same tags.

I had to remove redraw at every layer selection. Which lead to (1: removed resetWrap) and (2: dynamic CSS class).

src/js/controller/piskel/PublicPiskelController.js
```diff
@@ -17,7 +17,6 @@
     this.resetWrap_('setCurrentFrameIndex');
     this.resetWrap_('selectNextFrame');
     this.resetWrap_('selectPreviousFrame');
-    this.resetWrap_('setCurrentLayerIndex');
```
GMartigny: (1) see comment

src/js/controller/LayersListController.js
GMartigny: see comment

src/js/controller/LayersListController.js
```diff
@@ -123,14 +124,50 @@
     }
   };
 
-  ns.LayersListController.prototype.renameCurrentLayer_ = function () {
+  ns.LayersListController.prototype.focusLayerElement_ = function(element) {
+    var index = pskl.utils.Dom.getData(element, 'layerIndex');
+    this.piskelController.setCurrentLayerIndex(parseInt(index, 10));
+
+    var selectedClass = 'current-layer-item';
```
GMartigny: (2) See comment

src/js/utils/Event.js
```diff
@@ -7,6 +7,7 @@
     if (typeof el === 'string') {
       el = document.querySelector(el);
     }
+    scope = scope || this;
```
GMartigny: Ommiting scope throw an error. I prevent this behavior, but it could be wanted.

juliandescottes: We can throw a more explicit error, but I'd prefer to avoid this silent behaviour change.

src/js/controller/piskel/PublicPiskelController.js
juliandescottes: That's a good point but then we need to review why selectLayer should keep firing a RESET event. It's really just a different flavour of setCurrentLayerIndex.

Not against doing that but I would prefer if it was isolated in its own commit.

src/js/controller/LayersListController.js
juliandescottes: Even if we change the RESET event, the point of this comment was to say: there should be only one way to set the "current-layer-item" class on the selected element. 

Here the new code does it by DOM manipulation, while we used to set it by re-rendering the template before. I think we should remove the part of the template that can set this classname, isolate the code your new code to a dedicated method and also call it from renderLayersList_

src/js/controller/piskel/PublicPiskelController.js
GMartigny: When you say "its own commit", you mean a different PR ?

juliandescottes: We can do it in the same PR, having a dedicated commit will make it easier to review and to go back to. 

This commit should contain:
- removal of the reset wrapper for selectLayer 
- removal of the reset wrapper for setCurrentLayerIndex
- handling the selected layer classname by DOM manipulation instead of templating

Does that sound ok?

GMartigny: ok, got it.

TaniaRein:
Reviewed 3 of 5 files at r1, 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/piskelapp/piskel/633)*

TaniaRein:
Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[src/js/controller/LayersListController.js, line 118 at r2](https://reviewable.io/reviews/piskelapp/piskel/633#-Kc3tIsCmShteDeT6BXI-r2-118:-LCLwHcX1cRibirP93Bs:b-87j11e) ([raw file](https://github.com/piskelapp/piskel/blob/1de71bfd4c398ddf9cbe52e26a708f0683dacdf0/src/js/controller/LayersListController.js#L118)):*
<details><summary><i>Previously, GMartigny (Guillaume Martigny) wrote…</i></summary><blockquote>

see comment
</blockquote></details>

Good 

---


*Comments from [Reviewable](https://reviewable.io/reviews/piskelapp/piskel/633)*

TaniaRein:
Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[src/js/controller/LayersListController.js, line 128 at r2](https://reviewable.io/reviews/piskelapp/piskel/633#-Kc3tIsCmShteDeT6BXI-r2-128:-LCLwb3f-GVURDV25qyV:b-l0ge0b) ([raw file](https://github.com/piskelapp/piskel/blob/1de71bfd4c398ddf9cbe52e26a708f0683dacdf0/src/js/controller/LayersListController.js#L128)):*
<details><summary><i>Previously, juliandescottes (Julian Descottes) wrote…</i></summary><blockquote>

This method is creating new events every time it's called.
The events should either be cleaned up when we stop editing. 
Or only added once on the rootEl in the init. (would prefer that, it's more consistent with what we do in the codebase)
</blockquote></details>

Maybe maybe 

---


*Comments from [Reviewable](https://reviewable.io/reviews/piskelapp/piskel/633)*

TaniaRein:
Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[src/js/controller/LayersListController.js, line 140 at r2](https://reviewable.io/reviews/piskelapp/piskel/633#-Kc3tIsCmShteDeT6BXI-r2-140:-LCLwj7l8Kncfo7SOoNM:b32w1t) ([raw file](https://github.com/piskelapp/piskel/blob/1de71bfd4c398ddf9cbe52e26a708f0683dacdf0/src/js/controller/LayersListController.js#L140)):*
<details><summary><i>Previously, juliandescottes (Julian Descottes) wrote…</i></summary><blockquote>

See other comment about srcElement
</blockquote></details>

Ok

---


*Comments from [Reviewable](https://reviewable.io/reviews/piskelapp/piskel/633)*

juliandescottes:
Closing. @littlejd2010 please stop spamming the repository.

Issue #628, opened 2017-02-01
Two unknown characters in keyboard shortcuts cheatsheet


anijap:
*select previous frame* and *select next frame* have an unknown character next to them in the keyboard shortcuts cheatsheet window.

### Steps to reproduce the bug
1. go to piskelapp.com
2. click on *keyboard shortcuts* in the bottom left corner
3. *select previous frame* and *select next frame* have an unknown character next to them

### Environment details
* Operating System: Debian GNU/Linux 8 (jessie)  
* Kernel: Linux 4.8.0-0.bpo.2-amd64  

* browser (or offline application version): Firefox 45.7.0
(Current stable firefox version in Debian)

### Sprite details
N/A

![piskel_keyboard_missing_icons](https://cloud.githubusercontent.com/assets/4885345/22511484/b4125028-e84a-11e6-8816-4254471e43f7.png)

juliandescottes:
Thanks for the detailed bug report! 

(For the record the unknown keys are supposed to represesent the up and down arrows.)

The issue comes from CheatsheetController.js in the formatKey_ method.
https://github.com/juliandescottes/piskel/blob/master/src/js/controller/dialogs/CheatsheetController.js#L155-L158

Up/Down/Left/Right keys should be represented with the entities `&uarr;`, `&darr;`, `&larr;` and `&rarr;`.

PR #634, opened 2017-02-05
Use standard HTML entities for the arrows displayed on the keyboard shortcut list

zoeesilcock:
This is a fix for issue #628.

I don't have access to a Linux virtual machine at this point so it hasn't been tested. If someone has time to test it on Linux Firefox that would be great, otherwise I'll do it once I have access to a VM.

---
This change is [Reviewable](https://reviewable.io/reviews/juliandescottes/piskel/634)


juliandescottes:
That fixes it, thanks @zoeesilcock !

PR #759, opened 2017-10-23
Show drawing zoom level

### LINKED ISSUE #627: Show zoom level as text
It would be nice if the current zoom level was shown textually somewhere. Right now there's no way to tell when the main viewport is at e.g. 2x.

jordanjwatkins:
Option for displaying the current zoom level of the main drawing view (to resolve #627).

The zoom level would not fit nicely inline with the rest of the cursor coordinate info, so I landed on displaying it directly above.

### DIFF
--- a/src/css/layout.css
+++ b/src/css/layout.css
@@ -86,6 +86,12 @@
   font-family:monospace;
 }
 
+.cursor-coordinates .drawing-zoom {
+  position: absolute;
+  top: -20px;
+  left: 1px;
+}
+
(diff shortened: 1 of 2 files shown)


jordanjwatkins:
Demo: https://jordanjwatkins.github.io/piskel/

juliandescottes:
Thank you for the PR @jordanjwatkins ! 
Having a demo ready is really nice here. Merging immediately.

Issue #762, opened 2017-11-09
[Feature Request] Make better use of space in the user interface.


carlsmith:
I've been really getting into Piskel lately, and spent some time tweaking the UI. I just hacked it with Tampermonkey, but am interested in doing it properly and submitting a PR.

There's a screenshot below. I didn't really want the *Filename*, *Create Sprite*, *My Gallery* and *User* buttons at the bottom of the right-hand panel, but it was the best I could do with browser hacks. Ideally, they'd be in their own little titled section, just like *Layers*, *Transform* and *Palettes* (Maybe named *Account*), and I'd change the actual buttons to be exactly like the ones in the *Transform* section.

The main aim of the changes was just to make better use of the space, mainly getting rid of the banner across the top. I also made the background black, so that the black edge every physical panel has looks more like part of the UI, which creates a bit of space between the buttons and the edge of the screen. I also think black and gold imply a restricted palette, which feels appropriate.

If I made a PR, which changes would you like to see in it (if any)?

![screenshot](https://user-images.githubusercontent.com/7561668/32606844-b87cb7f2-c54e-11e7-9a65-09b569984ff2.png)

juliandescottes:
Thanks for the suggestion @carlsmith , that's interesting.

From the screenshot:
- I really like changing the coordinates / size color to gold, makes them popout a lot more
- we can make the background a bit darker to improve the contrast but I'd like to avoid going to pure black.
- not sure about moving the top bar to the right column. I already find it too packed as it is now :). The top bar also gives some consistency between the editor and the rest of the piskelapp.com pages. Maybe we could just have a "fullscreen mode" where the top bar gets collapsed completely  to the right, as a single menu button that would show on click the title of the sprite + links to all the actions currently available.

You can send a PR for the first items. For the top bar work, let's iterate a bit, because the code is split between this project and the piskelapp.com repo (https://github.com/piskelapp/piskel-website) so modifying this will be a bit more tricky :)

carlsmith:
Thanks for taking the time to look into it @juliandescottes - much appreciated. 

Changing the coordinates to gold is an easy edit, so I'll start with that. It'll let me get a PR done before looking at anything more involved.

Going to pure black would be a bit much. I wouldn't normally even consider it, and was just messing around with Tampermonkey to see how the 'hardware margin' would look, and was then surprised that it didn't look too bad. Still, it makes sense that you wouldn't want something that dark.

[...]

PR #764, opened 2017-11-11
Made cursor coordinates gold (and whitespace more consistent).

carlsmith:
As [discussed][1] with @juliandescottes, I've made the cursor coordinates at the bottom of the editor UI gold so they stand out better.

The CSS file (`layout.css`) generally has spaces after the colon between property names and their values, like `name: value`, but the rule this edit applies to didn't have the spaces, like `name:value`, so I added spaces to all four lines, but only otherwise edited one of them.

[1]: https://github.com/piskelapp/piskel/issues/762

### DIFF
--- a/src/css/layout.css
+++ b/src/css/layout.css
@@ -80,10 +80,10 @@
 }
 
 .cursor-coordinates {
-  color:#888;
-  font-size:12px;
-  font-weight:bold;
-  font-family:monospace;
+  color: gold;
+  font-size: 12px;
+  font-weight: bold;
+  font-family: monospace;
 }
 
 .cursor-coordinates .drawing-zoom {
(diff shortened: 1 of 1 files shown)


src/css/layout.css
```diff
@@ -80,10 +80,10 @@
 }
 
 .cursor-coordinates {
-  color:#888;
-  font-size:12px;
-  font-weight:bold;
-  font-family:monospace;
+  color: gold;
```
juliandescottes: Can we use var(--highlight-color) for gold? (defined in variables.css)

carlsmith: Sorry, didn't notice this till now. Of course. No problem.

juliandescottes:
Sorry I have been away from Piskel for the beginning of the week! Looks good, merging.

PR #782, opened 2018-02-16
Add a Grid Spacing option to the GridSettings

### LINKED ISSUE #774: [Feature Request] Grid Size
It would be real nice to have the option to choose the spacing between the grid, so you can work with tiles. 

At the moment it looks like it only allows you to change the width of the grid lines themselves but they are always 1px apart.

Craggar:
Copying styles/elements from the GridSize picker, this adds a GridSpacing picker

Fixes #774


juliandescottes:
Thanks for opening the PR! I'll have a look today.

src/css/settings-application.css
```diff
@@ -60,13 +60,15 @@
 }
 
 .settings-item-grid-size,
-.settings-item-grid-color {
+.settings-item-grid-spacing,
+.settings-item-grid-color,
```
juliandescottes: "," should be "{"

src/css/settings-application.css
```diff
@@ -60,13 +60,15 @@
...
   display: flex;
   align-items: center;
 }
 
 .settings-item-grid-size > label,
-.settings-item-grid-color > label {
+.settings-item-grid-spacing > label,
+.settings-item-grid-color > label,
```
juliandescottes: Same comment

src/js/utils/UserSettings.js
```diff
@@ -139,5 +141,10 @@
     if (storedGridWidth === 0) {
       ns.UserSettings.writeToLocalStorage_('GRID_WIDTH', 1);
     }
+
+    var storedGridSpacing = ns.UserSettings.readFromLocalStorage_('GRID_SPACING');
+    if (storedGridSpacing === 0) {
+      ns.UserSettings.writeToLocalStorage_('GRID_SPACING', 1);
```
juliandescottes: You don't need to add that, this was some migration code for version 12, because the GRID_WIDTH setting used to accept the "0" value. I should remove this method now :)

src/js/widgets/SpacingPicker.js
```diff
@@ -0,0 +1,51 @@
+(function () {
+  var ns = $.namespace('pskl.widgets');
+
+  ns.SpacingPicker = function (onChange) {
```
juliandescottes: I think we could reuse the SizePicker widget directly. The issue is mostly with the hardcoded "4" used in setSpacing. The goal was to represent the maximum value that can be represented via a [data-size] item. 

Instead of doing that we can change the logic to do:
- get the element for the provided size `this.container.querySelector('[data-size="' + size + '"]')`
- if the element is missing, get the last element and add a label:
```
      selectedOption = this.container.querySelector('[data-size]:last-child');
      selectedOption.classList.add('labeled');
      selectedOption.setAttribute('real-size', size);
```

With this small change we can reuse the SizePicker directly, just need to change the setSpacing to setSize and the data-spacing to data-size where appropriate.

juliandescottes:
Works really nicely, great job! 

One comment about the feature itself: it would be nice to change [computeGridWidthForDisplay_](https://github.com/piskelapp/piskel/blob/c2dbddcf9f94694f226092bcd1cb4b8c25318032/src/js/rendering/frame/FrameRenderer.js#L161-L167) so that it keeps displaying the grid a bit longer if grid spacing is enabled. If you use a a big spacing with a 320x320 sprite for instance, the grid disapears very quickly when zooming out.

The purpose of the computeGridWidthForDisplay_ method is to ensure that the grid doesn't become too dense or too thick.

Right now it does that with a magical `while (this.zoom < 6 * gridWidth) { gridWidth--; }`. This works well for thickness. But for density, if spacing is enabled we can keep displaying a thin grid on smaller zoom levels. Something like that would work:

```
  ns.FrameRenderer.prototype.computeGridWidthForDisplay_ = function () {
    var gridSpacing = this.getGridSpacing();
    if (this.zoom * gridSpacing < 6) {
      return 0;
    }

    var gridWidth = this.getGridWidth();
    while (gridWidth > 1 && this.zoom < 6 * gridWidth) {
      gridWidth--;
    }

    return gridWidth;
  };
```

Craggar:
Grand, thanks for the feedback - I should get to it today or tomorrow.

Craggar:
I've made the changes above, and removed the now-redundant SpacingPicker js/css to clean up the PR.

src/templates/settings/preferences/grid.html
```diff
@@ -19,6 +19,25 @@
...
+        <div class="size-picker-option"
+          title="4px" rel="tooltip" data-placement="top" data-size="4"></div>
+        <div class="size-picker-option"
+          title="8px" rel="tooltip" data-placement="top" data-size="8"></div>
+        <div class="size-picker-option"
+          title="16px" rel="tooltip" data-placement="top" data-size="16"></div>
+        <div class="size-picker-option"
+          title="32px" rel="tooltip" data-placement="top" data-size="32"></div>
+        <div class="size-picker-option"
+          title="64px" rel="tooltip" data-placement="top" data-size="64"></div>
```
juliandescottes: We still need to have custom sizes defined in CSS for those, otherwise the last four elements have the same size.

src/js/rendering/frame/FrameRenderer.js
```diff
@@ -298,12 +326,14 @@
...
     // Draw grid.
     var gridWidth = this.computeGridWidthForDisplay_();
+    var gridSpacing = this.getGridSpacing();
     if (gridWidth > 0) {
       var gridColor = this.getGridColor();
       // Scale out before drawing the grid.
       displayContext.scale(1 / z, 1 / z);
 
       var drawOrClear;
+      var drawing = true;
```
juliandescottes: variable is not used

PR #786, opened 2018-03-02
Use layer name as filename prefix in zip export

sharpfives:
Hi!

Piskel is such a great project; I use it all the time for making scenes and characters in my video games. Thanks for making this!

I'd like to have the exported layers optionally start with the layer name, rather than the layer index. For most of my games, I'm copying over the exported layers and frames from Piskel into another project, and then referencing them by filename in the game code. If I need to come back to Piskel and add another layer in the middle of others are re-export, my the existing layers' filenames are now shifted due to the index change and my game code's file references are wrong. Optionally using the layer name in the exported files rather than index would allow easier migration of files to other projects that aren't expecting layer file names to change each time.

I added a new checkbox to the zip export html template to enable this option, and supporting logic to the zip export controller. The new checkbox should not show unless the "split by layers" options is checked.

Let me know what you think.

Thanks again,
Alex

### DIFF
--- a/src/js/controller/settings/exportimage/ZipExportController.js
+++ b/src/js/controller/settings/exportimage/ZipExportController.js
@@ -15,16 +15,20 @@
     this.splitByLayersCheckbox = document.querySelector('.zip-split-layers-checkbox');
     this.addEventListener(this.splitByLayersCheckbox, 'change', this.onSplitLayersClick_);
 
-    this.useLayerNamesCheckboxContainer = document.querySelector('.use-layer-names-checkbox-container');
+    this.useLayerNamesContainer = document.querySelector('.use-layer-names-container');
     this.useLayerNamesCheckbox = document.querySelector('.zip-use-layer-names-checkbox');
-    this.useLayerNamesCheckboxContainer.hidden = !this.splitByLayersCheckbox.checked;
+    this.toggleHideUseLayerNamesCheckbox();
 
     var zipButton = document.querySelector('.zip-generate-button');
     this.addEventListener(zipButton, 'click', this.onZipButtonClick_);
   };
 
+  ns.ZipExportController.prototype.toggleHideUseLayerNamesCheckbox = function () {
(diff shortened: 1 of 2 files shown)


juliandescottes:
Thanks for the nice feedback and for the PR. The idea sounds good, I'll review the change this weekend.

src/js/controller/settings/exportimage/ZipExportController.js
```diff
@@ -12,12 +12,21 @@
     this.pngFilePrefixInput = document.querySelector('.zip-prefix-name');
     this.pngFilePrefixInput.value = 'sprite_';
 
-    this.splitByLayersCheckbox =  document.querySelector('.zip-split-layers-checkbox');
+    this.splitByLayersCheckbox = document.querySelector('.zip-split-layers-checkbox');
+    this.addEventListener(this.splitByLayersCheckbox, 'change', this.onSplitLayersClick_);
+
+    this.useLayerNamesCheckboxContainer = document.querySelector('.use-layer-names-checkbox-container');
+    this.useLayerNamesCheckbox = document.querySelector('.zip-use-layer-names-checkbox');
+    this.useLayerNamesCheckboxContainer.hidden = !this.splitByLayersCheckbox.checked;
```
juliandescottes: Could we hide this with CSS? Either set `style.display` set to `block` or `none`. Or `classList.toggle("hidden", !this.splitByLayersCheckbox.checked);`

juliandescottes: This could also get into its own method so that we can call it from here and from onSplitLayersClick_?

sharpfives:
Makes sense! I'll make those changes and push an update. Thanks.

juliandescottes:
Looks good, thanks for addressing the comments @sharpfives ! Merging. I'm preparing release v0.14 right now so your change should be available very soon on piskelapp.com :)

sharpfives:
@juliandescottes  No problem, thanks again for making this. Also nice work keeping the code clean and organized - made it super easy for me to get up and running.

PR #799, opened 2018-04-08
Jquery cleanup

juliandescottes:


juliandescottes:
This removes basic usage of jquery for:
- DOM apis
- function bindings

Jquery is still used for:
- tooltips 
- spectrum
- sortable

PR #819, opened 2018-05-22
Updated urls to use https

mbrickn:
Updated urls to use https instead of http when possible to avoid redirects.


juliandescottes:
Thanks @mbrickn !

PR #827, opened 2018-06-09
Resize single png export

### LINKED ISSUE #803: "Export selected frame" doesn't resize online
When choosing to "Export selected frame as PNG file", the scaling doesn't seem to work.
This is happening to me on Chrome, Windows 10, though it's probably happening everywhere.
Kind of a niche thing but still, it should work.

juliandescottes:
Fixes #803

### DIFF
--- a/src/css/settings-export.css
+++ b/src/css/settings-export.css
@@ -144,7 +144,8 @@
   overflow: hidden;
 }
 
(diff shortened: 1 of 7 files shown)


juliandescottes:
Why have I not merged that already?

PR #843, opened 2018-08-19
Fix grid bleeding through from other layers

### LINKED ISSUE #840: [Bug] Grid spacing is not updated until the next action
- Layers should be visible, 100% opacity
- 1 Frame, 1 Layer
- Go to PREFERENCES > Grid
- Enable Grid. Choose a grid spacing of eg 4px (make sure your current sprite size and zoom allows you to see the grid)
- Close PREFERENCES
- Paint the whole frame, color doesn't matter
- Create another Layer
- Go to PREFERENCES > Grid
- Change the Grid spacing to a bigger value

The smaller grid spacing is still visible (it remains on the first layer, for some reason).

icantsayrural:
By only enabling grid rendering for current layer. Fixes #840.

### DIFF
--- a/src/js/controller/DrawingController.js
+++ b/src/js/controller/DrawingController.js
@@ -22,7 +22,7 @@
 
     var cfg = {
       'zoom': this.calculateZoom_(),
-      'supportGridRendering' : true,
+      'supportGridRendering' : false,
       'height' : this.getContainerHeight_(),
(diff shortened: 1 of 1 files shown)


juliandescottes:
Thanks for figuring this out! Indeed we should only render the grid for the renderer that draws the frame.
Really nice to see contributors figuring this undocumented codebase :)

Issue #841, opened 2018-07-29
[Enhancement] Change opacity value to "alpha" icon 


juliandescottes:
The layer opacity currently shows "{value}α". 
I would like to try and replace this with an icon (maybe a big α ?) It could have three states:
- white: layer is fully opaque
- gray: layer has some transparency
- black (or black outline?): layer is fully transparent

This way the opacity "area" of the layer item would have a constant width, which is nicer if we want to stick another icon next to it.

ManzDev:
+1 for this

zakarh:
Hi there, is anyone currently working on this? I'd like to start working on this.

juliandescottes:
@zakarh assigning to you, thanks for your interest. Don't hesitate to post here if you have trouble getting started!

juliandescottes:
Ah I actually can't assign it to you directly, but feel free to get started! I will try to figure out why I can't assign it.

zakarh:
Awesome, thanks for the letting me know. I'll get started on this.

zakarh:
I added code so the alpha icon now has three opacity states. Also, I thought that numbers would better represent the opacity state so I didn't use colors. Here is a link to a video on [dropbox](https://www.dropbox.com/s/83qj0ek45qziq7r/Piskel%20Issue%20%23841.mp4?dl=0) where I demo this feature.
- 1α represents 100% opacity.
- 0.5α represents 50% opacity.
- 0α represents 0% opacity.

PR #845, opened 2018-09-03
Convert alpha value to an icon. Implementation for #841.

zakarh:
Implemented the enhancement for issue #841. Here is a [video](https://www.dropbox.com/s/83qj0ek45qziq7r/Piskel%20Issue%20%23841.mp4?dl=0) demoing the enhancement. Ran grunt test which completed without errors.

### DIFF
--- a/src/js/controller/LayersListController.js
+++ b/src/js/controller/LayersListController.js
@@ -143,7 +143,14 @@
     } else if (el.classList.contains('layer-item-opacity')) {
       index = pskl.utils.Dom.getData(el, 'layerIndex');
       var layer = this.piskelController.getLayerAt(parseInt(index, 10));
-      var opacity = window.prompt('Set layer opacity (value between 0 and 1)', layer.getOpacity());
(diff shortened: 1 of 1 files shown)


juliandescottes:
Thanks for the PR! However, this is not really what I had in mind. 

Users should still be able te decide which value they want for the opacity. But the representation of the opacity should just be an icon (therefore fixed-width) which has 3 states:
- `opacity == 1` -> icon is fully opaque 
- `0 < opacity < 1`-> icon is intermediary state
- `opacity == 0` -> icon looks "empty"

The first step here is to create this "3 states" icon.

zakarh:
Sure thing! So I'll modify the icon to have three states which use a color to represent the **opacity value set by the user**. The colors for each state as you suggested in #841 being...

- opacity == 1 -> icon is fully opaque, **color: white**
- 0 < opacity < 1-> icon is intermediary state, **color: gray**
- opacity == 0 -> icon looks "empty", **color: black (black outline)**

PR #846, opened 2018-09-03
Add frame toggling for preview

thorhunter1:
Hi, I've recently found out this project when looking for pixel editing software and it was one of the best free software I've found :) . But it lacks one feature that almost disqualified it from being chosen, namely toggling frames for animation preview. This is almost crucial for implementing several animations for one sprite.

Luckily, I've found out the project was open source and implementation of this feature quite possible and I think others may find it useful, so here it goes:

out

### DIFF
--- a/src/js/utils/serialization/Serializer.js
+++ b/src/js/utils/serialization/Serializer.js
@@ -21,7 +21,7 @@
       var serializedLayers = piskel.getLayers().map(function (l) {
         return pskl.utils.serialization.Serializer.serializeLayer(l);
       });
-      var frames = pskl.app.piskelController.getLayerAt(0).getFrames();
+      var frames = piskel.getLayerAt(0).getFrames();
       var hiddenFrames = frames.map(function (frame, index) {
         if (frame.visible) {
           return -1;
--- a/src/js/utils/serialization/Serializer.js
+++ b/src/js/utils/serialization/Serializer.js
@@ -21,6 +21,16 @@
       var serializedLayers = piskel.getLayers().map(function (l) {
         return pskl.utils.serialization.Serializer.serializeLayer(l);
       });
+      var frames = pskl.app.piskelController.getLayerAt(0).getFrames();
+      var hiddenFrames = frames.map(function (frame, index) {
+        if (frame.visible) {
+          return -1;
+        }
+        return index;
+      }).filter(function (frameIndex) {
+        return frameIndex !== -1;
+      });
+
       return JSON.stringify({
         modelVersion : Constants.MODEL_VERSION,
         piskel : {
@@ -29,7 +39,8 @@
           fps : pskl.app.piskelController.getFPS(),
           height : piskel.getHeight(),
           width : piskel.getWidth(),
-          layers : serializedLayers
+          layers : serializedLayers,
+          hiddenFrames : hiddenFrames,
         }
       });
     },
--- a/src/css/frames-list.css
+++ b/src/css/frames-list.css
@@ -136,7 +136,8 @@
 }
 
 .preview-tile .tile-overlay.tile-count.toggled {
-  background-color: rgba(111, 215, 0, 0.6);
+  background-color: gold;
+  color: black;
 }
 
 .preview-tile .tile-overlay.delete-frame-action {
--- a/src/css/frames-list.css
+++ b/src/css/frames-list.css
@@ -135,6 +135,10 @@
   cursor: default;
 }
 
+.preview-tile .tile-overlay.tile-count.toggled {
+  background-color: rgba(111, 215, 0, 0.6);
+}
+
 .preview-tile .tile-overlay.delete-frame-action {
    top: 0;
    right: 0;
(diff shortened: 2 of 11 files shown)


zakarh:
I really like this. When designing sprites animations I want to focus on specific frames. This makes it possible.

thorhunter1:
I'm glad you like it. After several bug fixes I started making actual sprites with it and it's very useful :).

juliandescottes:
Overall I really like this feature! Thanks a lot for the suggestion and for taking the time to make it work.

Regarding the implementation, we should be able to do something simpler, see suggestions below.

The "frame" objects are described in `js/model/Frame.js` . Here I suggest adding a `visible` property, as well as a toggleVisibility method. Then add a method on the piskelController (eg `toggleFrameAt`) that will toggle the visibility of all frames at the provided index. This is what should be called when performing the toggle action. Then you can remove all the state based on this.toggles in FramesListController. When rendering a frame you simply check if `frame.visible` is true, and set the class accordingly.

To render the animated preview, we can have a "hasVisibleFrameAt" method on the piskel controller and update the animation loop to do something like:

```javascript
      var index = Math.floor(this.elapsedTime / (1000 / this.fps));
      while (!this.piskelController.hasVisibleFrameAt(index)) {
        if (!this.piskelController.hasFrameAt(index)) {
          this.elapsedTime = 0;
          index = 0;
          break;
        }
        index++;
      }
```

I am not sure about the green color to show the frame as "toggled", but we can find another representation that fits the gold/black theme later :)

Another question is "should this apply when exporting" but this can also be handled later on.

thorhunter1:
Initially I took similar approach you proposed but it turns out the history management system is not very forgiving for those who want to add any data to frame model. Basically, any undo/redo causes frames to be recreated and I've given up on trying to add data into piskel state serialization/deserialization.

juliandescottes:
Ah! You probably had some issues with the ArrayBufferSerializer/Deseserializer. Happy to contribute that part of the code. I will try to push a commit that adds support for this, assuming the property and method names I mentioned above.

juliandescottes:
Here is a patch that should support both import/export and history service serializations:
https://gist.githubusercontent.com/juliandescottes/9639d34ecdd717afe79927c98b6d440d/raw/5e4f1786bc53220db2b0a572c8f177ebd651708c/gistfile1.txt

It only assumes a "visible" property on Frame. We can probably come up with cleaner code, but that should be enough to get started. Hopefully you still feel like giving this feature another shot :) Let me know if you have any trouble with the patch

juliandescottes:
Didn't think about it at first, but the other thing needed for the history service is to save the calls to this `toggleFrameAt`.

In PublicPiskelController, adding `this.saveWrap_('toggleFrameAt', true);` should automatically make it so that all calls to the method are part of the history. Meaning they are part of the undo/redo stack, but that's the easiest way to handle them here.

thorhunter1:
I followed your suggestions with the patch you provided and indeed everything seem to work nicely (except for unit tests :)). However, I kept the old implementation for calculating index for preview, because the one you suggested unfortunately has bugs (and I don't think it can be kept simple without running into corner cases). I've also changed color scheme to match gold standard :).

juliandescottes:
Thanks! I will take a look at the unit tests :)

src/js/utils/serialization/Serializer.js
```diff
@@ -21,6 +21,16 @@
       var serializedLayers = piskel.getLayers().map(function (l) {
         return pskl.utils.serialization.Serializer.serializeLayer(l);
       });
+      var frames = pskl.app.piskelController.getLayerAt(0).getFrames();
```
juliandescottes: This should be `var frames = piskel.getLayerAt(0).getFrames();` my bad!

juliandescottes:
My tests are failing locally, I must have busted my environment. The PR looks good, if you can just address my comment to make the unit tests pass, I will merge this.

juliandescottes:
Thanks! Merging

thorhunter1:
Ok coming back to this, it looks like there are some corner case bugs where history management toggles all frames instead of chosen ones. Also, creating new layer also doesn't have initial frames toggled properly so it looks like storing toggled status as part of state is not so simple after all :/

juliandescottes:
I'll take a look at this. 

HistoryManagement is operating in 2 modes:
- replay
- snapshot

Basically we store all actions in an array and when you undo we reapply all actions until the previous one.  To avoid redoing everything from the beginning, we regularly store a snapshot of the sprite (every 50 actions).

So when you hit CTRL+Z, we load the last snapshot taken, and then replay all actions. Maybe loading the snapshot fails to reapply the flag correctly, although that's precisely what the serializer work should have fixed.

The layer issue however seems very obvious now that you mention it.

Since our model is Piskel -> Layers -> Frames, the `visible` information becomes hard to maintain as soon as we have more than one layer. This means it should probably move the Piskel model instead, which should just keep the list of hidden indices.

Then we'll need to be careful about which methods also need to update this list (for instance when deleting a frame, moving a frame etc...). But in the long run that will probably be easier to handle than duplicating properly the `visible` information on Frames. 

Sorry for missing this when reviewing initially. Let me know if you want to try to address this, otherwise I'll take it.

PR #849, opened 2018-09-17
Add toggle grid icon to minimap

juliandescottes:


juliandescottes:
Should still:
- [ ] add integration test
- [ ] extract minimap actions to separate controller

juliandescottes:
Integration test done, also cleaned up the other tests to make them more stable.

juliandescottes:
Controller split (could be broken down further though) and added alt+G shortcut to toggle grid display.

PR #852, opened 2018-09-24
Add to frame size information about number of current frame

### LINKED ISSUE #771: Display current frame number
From twitter.

image

If you have many frames and the frame list doesn't display the current frame, you have no way to know which frame is currently selected. We should display the current frame number somewhere in the UI

janczer:
Fix #771 

Display current frame number.

display current frame number

### DIFF
--- a/src/js/controller/CursorCoordinatesController.js
+++ b/src/js/controller/CursorCoordinatesController.js
@@ -18,6 +18,7 @@
     $.subscribe(Events.DRAG_END, this.onDragEnd_.bind(this));
     $.subscribe(Events.FRAME_SIZE_CHANGED, this.redraw.bind(this));
     $.subscribe(Events.ZOOM_CHANGED, this.redraw.bind(this));
+    $.subscribe(Events.PISKEL_RESET, this.redraw.bind(this));
 
     this.redraw();
   };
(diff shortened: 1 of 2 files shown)


juliandescottes:
Thanks for jumping on this @janczer ! 

I think this should be in its own element rather than next to the size. If you start moving the cursor on the sprite you get a weird display such as 

image

I think we should aim for:
image

But the markup for this section is kind of a mess! 
In short we need to put each part of this display in its own element (`<div>s` are fine)
- zoom info
- size / cursor / selected area info
- frame index info

Then get rid of the "drawing-zoom" class (no idea why we are absolutely positioning this element, but that doesn't seem useful?). We may have to force text-align: left on the `.cursor-coordinates` container, other wise the column's `text-align: center` messes up everything. Finally increase a bit the line-height for this same container to make it look less packed.

Finally, there are some situations where the frame index can change (undo/redo for instance) which are not monitored by this controller. We should also redraw on `Events.PISKEL_RESET` in order to address this.

src/js/controller/CursorCoordinatesController.js
```diff
@@ -45,7 +46,13 @@
       html += '<div class="drawing-zoom">x' + zoom + '</div>';
     }
 
-    this.coordinatesContainer.innerHTML = this.getFrameSizeHTML_() + html;
+    this.coordinatesContainer.innerHTML = this.getFrameSizeHTML_() + html + this.getCurrentFrameIndex_();
+  };
+
+  ns.CursorCoordinatesController.prototype.getCurrentFrameIndex_ = function () {
```
juliandescottes: Can we just rename this method to have HTML in the name, for instance `getCurrentFrameIndexHTML_` (or shorter if you can find anything else :) _

janczer: Done.

janczer:
> Actually, I missed the huge package-lock.json, can we remove it ?

Of course, I'm sorry, it's was mistake. I reseted the latest commit and remove `package-lock.json`.

juliandescottes:
Looks great, thanks @janczer ! Will test it a last time later today and will merge afterwards :)

juliandescottes:
Forgot to merge, sorry :) This is working great, thanks again!

PrashBricks:
I am going to enjoy this feature.

Issue #172, opened 2014-04-24
Enhancement : Allow inline editing of layer's name


captainbrosset:
The old-looking prompt could probably be removed and instead an input could appear in place of the layer's name when clicking the edit icon.
I think that a dbl-click on the name should lead to the same thing.

![screenshot 2014-04-24 09 16 50](https://cloud.githubusercontent.com/assets/1152698/2786906/70074922-cb80-11e3-9c05-47fd2ea98ea6.png)

juanferrer:
Should this not be closed?

MischaDy:
As far as I can see this issue has been solved and can be closed. If not, please detail what is yet to be done. Thanks.

PR #857, opened 2018-10-08
Enabled inline editing of layer names

DoctorC0X:
Hi,

this is a pull request for issue https://github.com/piskelapp/piskel/issues/172
Let me know what you think!

Kind regards

### DIFF
--- a/src/css/toolbox-layers-list.css
+++ b/src/css/toolbox-layers-list.css
@@ -66,10 +66,15 @@
   cursor: pointer;
 }
 
-.layer-item .layer-name {
+.layer-item .layer-name,
+.layer-item .layer-name-input {
   padding: 0 0 0 10px;
   flex: 1 auto;
   white-space: nowrap;
+}
+
+.layer-item .layer-name-input {
+  width: 80%;
 }
 
 .layer-item .layer-name.overflowing-name {
@@ -82,7 +87,8 @@
 }
 
 .layer-item-opacity {
-  padding-right: 8px;
+  padding: 0 8px 0 8px;
+  flex: 0 auto;
 }
 
 .current-layer-item,
--- a/src/js/controller/LayersListController.js
+++ b/src/js/controller/LayersListController.js
@@ -3,11 +3,16 @@
 
   ns.LayersListController = function (piskelController) {
     this.piskelController = piskelController;
-    this.layerPreviewShortcut = pskl.service.keyboard.Shortcuts.MISC.LAYER_PREVIEW  ;
+    this.layerPreviewShortcut = pskl.service.keyboard.Shortcuts.MISC.LAYER_PREVIEW;
+    this.startRenamingCurrentLayer_ = this.startRenamingCurrentLayer_.bind(this);
+    this.onRenameInput_ = this.onRenameInput_.bind(this);
   };
 
   ns.LayersListController.prototype.init = function () {
+    this.isRenaming = false;
+
     this.layerItemTemplate_ = pskl.utils.Template.get('layer-item-template');
+    this.layerNameInputTemplate_ = pskl.utils.Template.get('layer-name-input-template');
     this.rootEl = document.querySelector('.layers-list-container');
     this.layersListEl = document.querySelector('.layers-list');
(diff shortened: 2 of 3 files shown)


DoctorC0X:
#172

src/js/controller/LayersListController.js
```diff
@@ -130,6 +133,18 @@
...
       layerNameEl.setAttribute('rel', 'tooltip');
     }
+    if (isRenaming) {
+      var layerNameInputHtml = pskl.utils.Template.replace(this.layerNameInputTemplate_, {
+        'layername': layer.getName()
+      });
+      var layerNameInput = pskl.utils.Template.createFromHTML(layerNameInputHtml);
+      var layerNameEl = layerItem.querySelector('.layer-name');
+      layerItem.replaceChild(layerNameInput, layerNameEl);
+      layerNameInput.addEventListener('blur', this.onRenameCurrentyLayerDone_.bind(this));
```
juliandescottes: Here we will be leaking these two event listeners every time we create this input. In `ns.LayersListController.prototype.init` it is ok to do it because we will only call init() once in the whole lifecycle of the application. 

Here the method is called every time we create the input so we have to be more careful.
You can either use event delegation or remove the events when setting `this.isRenaming` to `false`. I think the second approach will be easier so a few pointers:
- bind the method once and for all in the constructor: `this.onRenameCurrentyLayerDone_ = this.onRenameCurrentyLayerDone_.bind(this)`
- attach your listeners without calling bind() again `layerNameInput.addEventListener('blur', this.onRenameCurrentyLayerDone_)`
- remove the event listeners right before setting this.isRenaming to false
- to be safe, also remove the event listeners right before adding them (this way if addLayerItem is called twice for some reason, the events will only be added once)

src/css/toolbox-layers-list.css
```diff
@@ -53,13 +53,14 @@
  */
 
 .layers-list {
+  display: table;
```
juliandescottes: This breaks the scroll if there are too many layers. 

Do we really need to change the layout from flex to table? What was the issue with the current flex approach?

src/js/controller/LayersListController.js
```diff
@@ -46,26 +49,26 @@
 
   ns.LayersListController.prototype.createButtonTooltips_ = function () {
     var addTooltip = pskl.utils.TooltipFormatter.format('Create a layer', null, [
-      {key : 'shift', description : 'Clone current layer'}
+      { key: 'shift', description: 'Clone current layer' }
```
juliandescottes: Most of the codebase has a space before the colon in objects. 
It would be nice to preserve this in the patch, especially on lines which were not touched by the patch otherwise.

src/js/controller/LayersListController.js
```diff
@@ -157,13 +172,22 @@
...
+
+  ns.LayersListController.prototype.onRenameCurrentyLayerDone_ = function (evt) {
+    var el = evt.target || evt.srcElement;
+    if (evt.key && evt.key !== 'Enter' && evt.key !== 'Escape') {
+      return;
+    }
+    var name = el.value;
     if (name) {
       var index = this.piskelController.getCurrentLayerIndex();
       this.piskelController.renameLayerAt(index, name);
```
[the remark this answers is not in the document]
DoctorC0X: I agree. Blurring and pressing escape will now abort the renaming.

src/js/controller/LayersListController.js
DoctorC0X: Very nice feedback, thanks. Done.

src/css/toolbox-layers-list.css
DoctorC0X: I had some problems with the alpha (opacity) symbol to stay in place when starting the renaming. So I thought a table would be easier to manage. But I didn't notice that I broke the scroll, good catch ;) . I now use the flex again, let me know what you think, as I'm not so proficient with css.

juliandescottes:
Will test that right now, thanks for the update!

juliandescottes:
Looks good and the patch works nicely, thanks a lot!

PR #860, opened 2018-10-18
Add piskel-cli for exporting from .piskel files via command line

### LINKED ISSUE #853: [Feature request] Command line argument to do the export programatically
First of all, thank you for this great app !

I would like to (try to) make a game and edit sprites with the desktop version of piskel, and it would be very nice for me (and perhaps others) to be able to do the export programatically.

In fact, I'd like to be able to put the .piskel files on git so I can get them and edit them from everywhere, and avoid one of the two following things:

  - put also the png files on the git
  - manually export all sprites

Having a command line tool to do the export would allow me to generate the image files just before compiling for example.

jordanjwatkins:
This aims to resolve #853.

There's still plenty to do for full feature parity with the app's exporting, but I wanted to keep the initial commit small and manageable for review.

This iteration handles most of the PNG export options available via the UI, except for PixiJS Movie export (though I have it working in another branch).

It also includes single frame PNG export from the 'Others' export options and basic cropping.

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -1,0 +1,64 @@
+# Piskel CLI
+
+Wraps the Piskel pixel editing application to enable similar export options via the command line.
+
+## Installation
+
+Option 1: Globally install Piskel
+```
+npm install -g https://github.com/piskelapp/piskel/tarball/master
+```
+
+Option 2: Clone and install Piskel normally and then run npm link inside the installation root
+
+## Usage
+
+**Export provided .piskel file as a png sprite sheet using app defaults**
+```
+piskel-cli snow-monster.piskel
+```
+
+**Export scaled sprite sheet**
+```
+piskel-cli snow-monster.piskel --scale 5
+```
+
+**Export scaled to specific (single frame) width value**
+```
+piskel-cli snow-monster.piskel --scaledWidth 435
+```
+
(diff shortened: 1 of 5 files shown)


juliandescottes:
When you opened the issue, I assumed the easiest would probably be to duplicate the deserializer logic and do this in isolation, without Piskel files. As you probably realised the "module system" (if we can call it that) of Piskel is almost impossible to reuse somewhere else. The only solution is what you did (or variants of it). Congratulations for wiring that up :)

I have no issue with merging this, since it doesn't impact Piskel itself. But would you like to try to extract this in its own repository? This way you can keep working on it and releasing it without having to wait on me for reviews and merges. 

We can merge for now, and work on extracting your cli later. As you want

jordanjwatkins:
I removed the package-lock.json file.

I also switched data-uri rendering methods to use the version provided by phantom since toDataURL was creating a much larger string when executed in the phantom page context for some reason.

This version does support providing an alternate name and output folder via the 'dest' argument, but I hadn't included it in the usage documentation, so I updated that.

I was already toying with the idea of a separate package for the CLI and tried it out, but adding to the existing repo seemed like the smallest iteration and somewhere to get feedback.

If you're on the same page, I probably will go ahead and publish it as a separate repo and npm package when I get a chance to keep tinkering on it.

PR #862, opened 2018-10-19
Add option to export CSS for a spritesheet

### LINKED ISSUE #542: CSS for spritesheet export
Hi,

With the new spritesheet export and spritesheet layout configuration, it could be nice to have the corresponding CSS to access each sprite.
Because today, I have to export as zip -> unzip -> run a sprite-smith.

It could be a very basic list of rules like :

    .sprite-layerName {
        background: url(piskelName.png) -32px -16px;
    }

Basic LESS could also be nice to have.

jordanjwatkins:
Feature to resolve #542.

Keeping it simple and only generating CSS for now (no LESS).


jordanjwatkins:
Oops, my fork was not fresh, but the code and demo are up-to-date now.

src/templates/settings/export/png.html
```diff
@@ -33,6 +34,16 @@
       </div>
     </div>
 
+    <div class="export-panel-section">
```
juliandescottes: For me it doesn't fit if I have more than one frame in my sprite. When there is more than 1 frame there is an extra "layout section" that is introduced in the `PNG` tab.

Since this doesn't really care about the layout options, we could move it to the `Other` tab?

juliandescottes: Hmmm actually it does use the layout options.

jordanjwatkins: So should it go in the 'other' tab or should I bump up the height of that drawer?

jordanjwatkins: With a multi-frame sprite I can get the drawer to (just) fit without any overlap on my pretty small laptop screen... but it's quite snug.

Initially, I had the CSS export as a checkbox option with the sprite sheet export which was slightly less height intensive, so that's maybe another option.

Let me know where you want it and I'll adjust and/or move it.

Brandonbr1:
I Think the export to css should be in the others tab not in the PNG tab

PR #906, opened 2019-11-11
add grunt-cli to devDependencies, fix nwjs start script, add serve and build commands

blurymind:
You cant build piskel without it, yet when you npm install, it's not installed. This kind of breaks the usual npm install, then run approach as the user has to then figure out why piskel is not building and what to do

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   "main": "./dest/prod/index.html",
   "scripts": {
     "test": "grunt test",
-    "start": "nodewebkit",
+    "start": "nw",
     "preversion": "grunt test build",
     "postversion": "git push && git push --tags && npm publish",
     "release": "grunt && node ./bin/copy-to-piskel-website"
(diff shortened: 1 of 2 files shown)


blurymind:
this pr also fixes the start script, which is currently broken

blurymind:
It also adds the commands from the wiki to serve and build the app, so people can look at package.json and do it right away

package.json
```diff
@@ -23,16 +23,21 @@
   "main": "./dest/prod/index.html",
   "scripts": {
     "test": "grunt test",
-    "start": "nodewebkit",
+    "dev": "grunt serve",
```
juliandescottes: Note: there is also the badly named `grunt play` which starts a dev server with "debug" files (unbundled, unminified). Since piskel doesn't have sourcemaps or anything, it can be quite useful.

blurymind: perhaps we should start it instead of grunt serve, if it starts faster?

"dev": "grunt play"

juliandescottes: It is slightly faster to start. And being able to debug real sources, get readable stacktraces etc. feels more developer friendly to me. But I am also ok with adding a second script ("dev-debug"?) later on.

As you prefer, just let me know what you decide and I'll merge this!

blurymind: lets make it use grunt play instead :) it does start faster and is better for debugging. You're right!

juliandescottes:
Thanks @blurymind ! 
Had a busy week, will get to the other PRs today.

PR #908, opened 2019-11-11
Add ability to source all layers to bucket fill tool

blurymind:
This makes the bucket fill tool use all layers as reference for what area to fill, not just the one you are on. This is made the default.

The ability to also source only current layer (old behaviour) is preserved and assigned to a modifier key (ctrl)

### DIFF
--- a/src/js/tools/drawing/PaintBucket.js
+++ b/src/js/tools/drawing/PaintBucket.js
@@ -10,6 +10,8 @@
     this.toolId = 'tool-paint-bucket';
     this.helpText = 'Paint bucket tool';
     this.shortcut = pskl.service.keyboard.Shortcuts.TOOL.PAINT_BUCKET;
+
+    this.tooltipDescriptors = [{ key: 'ctrl', description: 'Source only current layer' }];
   };
 
   pskl.utils.inherit(ns.PaintBucket, ns.BaseTool);
@@ -19,7 +21,15 @@
    */
   ns.PaintBucket.prototype.applyToolAt = function(col, row, frame, overlay, event) {
     var color = this.getToolColor();
-    pskl.PixelUtils.paintSimilarConnectedPixelsFromFrame(frame, col, row, color);
+    
+    var sourceOnlyCurrentLayer = pskl.utils.UserAgent.isMac ? event.metaKey : event.ctrlKey;
+    var sourceFrame = frame;
+    if (!sourceOnlyCurrentLayer) {
+      var currentFrameIndex = pskl.app.piskelController.getCurrentFrameIndex();
+      sourceFrame = pskl.utils.LayerUtils.mergeFrameAt(pskl.app.piskelController.getLayers(), currentFrameIndex);
+    };
+
+    pskl.PixelUtils.paintSimilarConnectedPixelsFromFrame(sourceFrame, frame, col, row, color);
 
     this.raiseSaveStateEvent({
       col : col,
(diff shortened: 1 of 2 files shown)


blurymind:
Hi Thank you for having a look at this :) 
 Would you be ok to have it assigned to the modifier then? 

So what you are doing there is clicking on a filled area on the layer below. That is why nothing happens - you are filling with black an area that is already black.
This is what happens in other software too, but it is not entirely obvious in piskel, because it applies a transparency effect when displaying layers that you are not on.

Perhaps the display should not have that effect when you are using the bucket tool, so it doesnt look like something is wrong with the tool.

The way the tool works in all other software is:
1. get the pixels from all visible layers and combine them into one layer
2. source that combination and use it to generate a fill area result
3. draw the resulting pixels on the currently active layer

juliandescottes:
Hi @blurymind, thanks for the added info!

> This is what happens in other software too
I would be interested to look at examples. I checked GIMP and Aseprite yesterday, they don't work like that. I don't seem to remember that photoshop or paint.net handle the fill/paint bucket tools in this manner either?

> Would you be ok to have it assigned to the modifier then?
If I can look at prior art in other applications I will consider it. That's the only way I can gauge if this is somewhat usual for graphical tools. 

Thanks!

blurymind:
@juliandescottes  sorry I had to mention that in gimp, in order to get this behaviour, you have to enable "Sample merged"

To be clear, this feature is called Bucket tool - sample merged mode
sampleMergedGimp

I believe mypaint does it the same way - has the same option

juliandescottes:
Thanks! 

In the PR the behavior of the tool is not the same as in GIMP. When I clicked in the transparent area above the square from the below layer, then it should have filled the pixels of my current layer in black, in the area corresponding to the connected pixels from the "merged" frame.

Now that I have another implementation example, I will take another look. The issue that will remain is how much we should expose this. We can't possibly add all the options of more advanced tools as modifiers :) I think if we want to open the door to more tool settings, we'll need to have an actual UI to select tool settings

blurymind:
> Now that I have another implementation example, I will take another look. The issue that will remain is how much we should expose this. We can't possibly add all the options of more advanced tools as modifiers :) I think if we want to open the door to more tool settings, we'll need to have an actual UI to select tool settings

Hi yes I agree the default behaviour should not be what is implemented, but the way it was before. The implementation should be on a modifier instead.

Ideally I would like it to be a tick box like in gimp, but  the current UI doesnt allow for checkbox style tool modifiers and I didn't want to risk making the pr harder to take in by introducing new ui concepts.

Would you be ok with allowing the new behaviour as a modifier, under the condition that it works the same as gimp with "sample merged" enabled? Then once we have some sort  of infrastructure to add tool inputs, I will do another pr to turn it into a checkbox rather than a modifier?

blurymind:
I added a gif on an example use case where this bucket fill mode is needed:
69063084-c55bb700-0a13-11ea-9516-caa94a78b312

The ability to sepparate your fill layer from your lineart layer!

blurymind:
I need to update this to work exactly like Gimp, but there will be no point in doing so if it cant be assigned to a modifier

juliandescottes:
> I need to update this to work exactly like Gimp, but there will be no point in doing so if it cant be assigned to a modifier

Right, sorry for not giving a clear answer. Let's go for a fixed version of the tool, behind a modifier. The last example you shared seems useful for keeping the lineart separated from the rest.

blurymind:
In that case I will look into making it behave exactly like gimp's and update this pr

PR #909, opened 2019-11-16
Add color palette apply tool

blurymind:
This adds a new tool to the transformsection, that allowsyou to apply the currently selected palette to all the colors of the frame or the colors of all frames..

The pr also adds a new dynamic palette, called "Frame colors". This palette automatically updates itself to show you all the colors of the selected frame and their index numbers. Those index numbers are what the new tool uses for reference


blurymind:
For anyone who wants to try all of the new features I am adding in one branch, check out
https://github.com/blurymind/piskel/tree/piskel-plus

Mrguy004:
Can add a tool can made  color Transparent?

blurymind:
Possible to do :) seems like a corner case type of use - perhaps a mode for the eraser tool?
Colors in piskel dont seem to have an alpha value in the palette editor

juliandescottes:
> Can add a tool can made color Transparent?

Not sure if this is what you are looking for, but in the current version there is the "special" paint tool which can be used to change the color of all the pixels that share the same initial color.

image

image

If you use it with the transparent color, it will effectively erase all the pixels of a given color. By default only applies to current frame / current layer, but can be applied to all frames / all layers with `ctrl` & `shift`.

blurymind:
I would suggest to do what @juliandescottes recommends.If the tool doesn't do what you want, raise a separate issue for it. This pr is for colour palette swapping feature and anything outside of that is out of its scope

blurymind:
To be clear, this new tool is a Transform type operation for palette swapping. The way it works is
69063093-cb519800-0a13-11ea-93ea-2024189519c1

You have a dynamic reference palette (current frame palette), which gets updated every time you add a new color to your frame. You can then duplicate your reference palette and change any of the colours in it. Select that new palette (should be selected when you create it), then using this tool, the colours in it will be compared to the reference palette. A colour replace operation is applied to each of the reference colours, using their indexes - the same operation that the tool @juliandescottes showed uses - but automatically on each colour of the frame.
Howding a modifier key will apply the operation on all frames as swapout all the colours with the ones  you have in your new palette - in a single click

PR #934, opened 2020-05-05
Add "Outline tool" for quickly outlining shapes.

ElectricToy:
The Outline Tool works similarly to the paint bucket, but only paints pixels along the edges of shapes. Hold Ctrl/Cmd to have it fill corners too.

I didn't create a tool icon. I copied-and-pasted the paint bucket icon as a stand-in, but I'm not so sure I did that correctly.

The tool works quite nicely, however. Try it!


ElectricToy:
I ran the Travis-CL checks locally and they pass. The log above indicates that Phantom crashed, apparently of its own accord.

gingerbeardman:
Thanks so much for this! I use it in my own custom build.
https://github.com/gingerbeardman/piskel-playdate/blob/dev-1047/README.md

Issue #980, opened 2020-12-18
Difficult Frame Selection With Graphics Tablet


rkhoury:
Bug description

Selecting a frame from the left-side container when using a graphics tablet is extremely difficult as most mouse events appear to transition to drag event rather than click.

I'm not at all familiar with JS coding but a quick look around FramesListController.js and some of the logic regarding selection and drag, it looks as though the behavior I've described is intended ("click" event listening + use of `justDropped` value). 

I was looking to modify the code such that frame selection happens on mouseDown rather than "click" (in which case a drag even will always result in selection of the dragged frame, which I thought would be okay?).
If that's not desirable, is it possible to just allow a tile select operation if the drag event didn't move more than X(a small) screen units?

### Steps to reproduce the bug
1. go to piskelapp.com
2. Create more than one frame
3. Using the pen from a graphics tablet attempt to change the frame selection.

### Environment details
* Windows or Linux
* Chrome or Firefox or Linux desktop.

### Sprite details
* number of frames: > 1
* sprite resolution: N/A
* session duration: N/A

PR #982, opened 2021-01-01
Making sure frame selection via graphics tablet works.

rkhoury:
Frame selection with a graphics tablet often results in a drag/drop
event and so selection didn't occur very easily. This change ensures
that the drag event will result in a selection always.

Relating to issue #980.


rkhoury:
Closed this since the latest doesn't seem to suffer from the same issue I had experienced.

PR #983, opened 2021-01-02
Further customization of grid spacing

Nickvs99:
Currently, the grid spacing is limited to a set number of widths: 1, 2, 4, 8, 16, 32, or 64 pixels. The grid spacing with these proposed changes can be any integer value. Thus the user is not restricted in his grid spacing, but is free to choose its own. The proposed changes can be seen on the left, while the original can be seen on the right of the image.

<a href="https://imgur.com/9dg8jWu"></a>

### DIFF
--- a/src/css/settings-application.css
+++ b/src/css/settings-application.css
@@ -73,26 +73,9 @@
   flex-shrink: 0;
 }
 
-.grid-spacing-container .size-picker-option[data-size='1'] {
-  padding: 7px;
-}
-.grid-spacing-container .size-picker-option[data-size='2'] {
-  padding: 6px;
-}
-.grid-spacing-container .size-picker-option[data-size='4'] {
-  padding: 5px;
-}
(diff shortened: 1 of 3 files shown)


OPStellar:
Would it also be possible to add a custom swatch for the grid color?

Nickvs99:
Yes, it is now possible to change the color of the grid to any possible color. The grid menu now looks like the image on the left and when the color selector is clicked it expands into a color swatch (right image).

<a href="https://imgur.com/th9SYpx"></a>

OPStellar:
It just occurred to me that maybe the grid spacing could have an up-and-down arrow by it, so it could be adjusted incrementally. But then if it were to have that, it would need to be implemented in other places.
Anyway, thanks for incorporating the feature!

UnderdogeHeroine:
Bump please add 48x48 pixels, thats pretty much standard for many a game design application <3

Lt-knb:
Why does the web app look the same?
Was this ever merged?

PR #1099, opened 2023-05-10
added frames counter when importing them

JeffVenancius:
I've always had to calculate how many pixels I would need to separate the spritesheet accordingly when importing, so I've made inputs to place by frame instead of size.

You know: Instead of setting 32 x 32 for a spritesheet with 4 x 4 sprites, you just set 4 x 4 on this new area and it will do the rest.


gingerbeardman:
I did something similar, but in my situation the cell size in encoded in the filename (Playdate image table) so I could parse that and set the values.

JeffVenancius:
> Nice! I did something similar, but in my situation the cell size in encoded in the filename (Playdate image table) so I could parse that and set the values.

Cool! Playdate looks very interesting, hope it's doing well.

PR #1235, opened 2026-01-08
Export animation as apng format

cangSDARM:
I dunno if it’s right to stick the APNG export option under "Others", but yeah, we can export as apng now!

using: https://www.npmjs.com/package/@pdf-lib/upng

### DIFF
--- a/src/piskel-script-list.js
+++ b/src/piskel-script-list.js
@@ -56,7 +56,7 @@
   // JSZip https://github.com/Stuk/jszip
   "js/lib/jszip/jszip.min.js",
 
-  // @pdf-lib/upng https://github.com/Hopding/upng
+  // upng-js https://github.com/photopea/UPNG.js
   'js/lib/apng/UPNG.js',
 
   "js/lib/scrollifneeded/scrollifneeded.js",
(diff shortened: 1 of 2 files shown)


src/piskel-script-list.js
```diff
@@ -204,7 +207,7 @@
   "js/service/keyboard/Shortcuts.js",
   "js/service/keyboard/ShortcutService.js",
   "js/service/ImportService.js",
-  "js/service/ImageUploadService.js",
+  // "js/service/ImageUploadService.js",
```
cangSDARM: This file isn’t in the repo. I wonder if this is a bug.

ej-sanmartin:
Could go under PNG but under Other seems reasonable as well. What's the APNG use case?

src/piskel-script-list.js
ej-sanmartin: I believe this file no longer exists anymore. In my fork of piskel, https://github.com/Unveil-gg/RetroPiskel, i caught this as well and just removed it. Would be interesting to see in the past what ImageUploadService.js was used for.

PR #1237, opened 2026-01-12
import spritesheet with margin

cangSDARM:
I’m not sure if I should call it margin, padding, or gap. But hey, we now support importing with spaces between frames!

P.S. I haven’t tested the content yet—might need someone to help fill in the blanks :)

### DIFF
--- a/src/css/forms.css
+++ b/src/css/forms.css
@@ -14,6 +14,10 @@
   height: 23px;
 }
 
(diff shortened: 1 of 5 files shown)


cangSDARM:
you can visit my hosted version at https://cangSDARM.github.io/piskel, includes all my changes.

PR #1266, opened 2026-04-18
IE11 cleanup

grosbouddha:
Remove IE11 specific code:
 - IE11 worker code: for IE11, we were serving an actual file instead of using a in-memory blob url (required for ES module migration)
 - Drop obsolete `navigator.msSaveBlob` and  `window.saveAs` in downloadAsFile
 - various other small IE11 specific hack/branches
 - Remove IE11 from list of supported user agents
  
 TESTED=Locally by downloading a file, e2e passing


juliandescottes (approved):
Could potentially remove 

https://github.com/piskelapp/piskel/blob/a6b9c02daefceb10093f71e92d52d16920ccb16e/src/css/reset.css#L33-L36

as well, or just drop the comment if we are worried about visual regressions.

PR #1267, opened 2026-04-18
Remove Q.js lib

grosbouddha:
Drop the [q.js](https://github.com/kriskowal/q) promise library in favor of native Promise:

- Q.defer() → new Promise((resolve, reject) => ...)
- Q.resolve / Q.reject → Promise.resolve / Promise.reject
- Q.all([...]) → Promise.all([...]) 
- Q([]) → Promise.resolve([])

No Q-specific features were used anywhere in app code, so the migration is pure 1:1 with no behavior changes.

The trickiest part was ensuring proper this binding where needed - constructing a Promise requires an anonymous function that shadows the enclosing this.

TESTED=Unit tests and E2E passing


src/js/service/storage/GalleryStorageService.js
juliandescottes, line 1: I think the whole file can be deleted, we can drop it early rather than carrying it along for the refactors?

.husky/pre-commit
```diff
@@ -1 +1 @@
-npx biome check --staged --write --no-errors-on-unmatched
+npm run lint
```
juliandescottes, line 1: Just checking if this is intentional.

Follow up: Noticed we are only linting on src/ 
We should extend to tests/ no reason not to lint and format test code.

juliandescottes (approved):
Thanks! Now that looks like JS from 2017 or something :)

src/js/service/storage/GalleryStorageService.js
juliandescottes: Alright I see this is the next PR 👍

PR #1268, opened 2026-04-18
Dead code cleanup

grosbouddha:
Remove several layers of dead code and obsolete feature remnants. No behavior change for any supported browser:

- Delete empty `src/js/snippets.js`
- Drop `Function.prototype.bind` and `Uint32Array.prototype.fill` polyfills; both native in every browser
- Remove gallery stuff*: `pskl.app.isLoggedIn()` always returned `false`. Remove it plus the dead chain it gated: `GalleryStorageService`, `StorageService.saveToGallery`, the `.save-public-section` UI, `SaveController.isPublic_()`, and `Descriptor.isPublic
- Remove last bits of app engine integration
- Delete `RELEASE.md` (described a Google AppEngine Launcher workflow no longer in use and the rest is obsolete or incorrect) 

NOTE: I excluded tests from the husky precommit (tests are not passing the lint) by using "npm run lint" in the precommit hook (which only runs on src/ for now).


.husky/pre-commit
```diff
@@ -1 +1 @@
-npx biome check --staged --write --no-errors-on-unmatched
+npm run lint
```
juliandescottes, line 1: Same comment as on the other PR, is this intentional?

juliandescottes (approved):
Thanks for the cleanup

